Issue #11, opened 2020-03-29
Implement a generic libvirt inventory plugin


odyssey4me:
##### SUMMARY

It would be very useful to have an inventory plugin for libvirt which works against both local and remote libvirt URI's. It can produce inventory which designates the host, groups according to type (lxc, qemu, etc) and provides various details like IP addresses and such.

##### ISSUE TYPE
- Feature Idea

##### ADDITIONAL INFORMATION

```yaml

```

Issue #10, opened 2020-03-29
Convert scripts/inventory/libvirt_lxc.py into an inventory plugin


odyssey4me:
##### SUMMARY

Dynamic inventory is better served through the use of the inventory plugin interface. Convert the script into an inventory plugin and ensure that it has unit and test coverage.

##### ISSUE TYPE
- Feature Idea

##### ADDITIONAL INFORMATION

```yaml

```

PR #23, opened 2020-04-10
Add generic libvirt inventory source

daveol:
##### SUMMARY
Working on implementing an generic libvirt inventory plugin that uses the python api

Fixes: #10 
Fixes: #11 

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
inventory/libvirt

##### ADDITIONAL INFORMATION

```paste below

```


odyssey4me:
Hi there - I'm very happy that you're picking this up. Thank you! Note that https://github.com/ansible-collections/libvirt/issues/12 is totally different, and I think is better left to another PR. I've actually done some of the work for it and will push up a draft PR shortly.

odyssey4me:
FYI I've pushed up an initial pass at a connection plugin in https://github.com/ansible-collections/libvirt/pull/26 - reviews/suggestions would be appreciated!

plugins/inventory/libvirt.py
```diff
@@ -0,0 +1,69 @@
...
+    - constructed
+short_description: Libvirt inventory source
+description:
+    - Get libvirt guests in an inventory source
+author:
+    - Dave Olsthoorn <dave@bewaar.me>
+version_added: "2.10"
+options:
+    plugin:
+        description: token that ensures this is a source file for the 'libvirt' plugin.
```
odyssey4me, line 17: nit: It'd be nice if 'token' could be capitalised as 'Token' given it's the start of the sentence.

plugins/inventory/libvirt.py
```diff
@@ -0,0 +1,69 @@
...
+        required: True
+        choices: ['libvirt']
+    connection:
+        description: Libvirt Connection URI
+        required: True
+        type: str
```
odyssey4me, line 23: I would like this if it could be a list, so that it could gather from multiple hosts instead of just one. That could be a follow on bit of work though.

plugins/inventory/libvirt.py
```diff
@@ -0,0 +1,69 @@
...
+        super(InventoryModule, self).parse(
+            inventory,
+            loader,
+            path,
+            cache=cache
+        )
+
+        conn = libvirt.open(self.get_option('connection'))
+        if not conn:
+            raise Exception("hypervisor connection failure")
```
odyssey4me, line 55: Exceptions should be raised via AnsibleError, so with

from ansible.errors import AnsibleError

you can raise the error like so:

raise AnsibleError("hypervisor connection failure")

plugins/inventory/libvirt.py
daveol: that should already be possible with [multiple inventories](https://docs.ansible.com/ansible/latest/user_guide/intro_inventory.html#using-multiple-inventory-sources) tough... so should there be explicit support for this?

odyssey4me: Oh! good point! That totally makes sense. Never mind then. :)

plugins/inventory/libvirt.py
daveol: Done

odyssey4me:
LGTM, I think there's quite a lot of improvement to do for functional testing - but that's a general issue.

Issue #12, opened 2020-03-29
Implement a generic libvirt connection plugin


odyssey4me:
##### SUMMARY

It would be very useful to have an connection plugin for libvirt which works against both local and remote libvirt URI's without requiring access to the machine via SSH. This feature can make use of the libvirt guest agent.

References:
- https://wiki.libvirt.org/page/Qemu_guest_agent
- https://wiki.qemu.org/Features/GuestAgent
- https://qemu.weilnetz.de/doc/qemu-qmp-ref.html

##### ISSUE TYPE
- Feature Idea

##### ADDITIONAL INFORMATION

```yaml

```

PR #26, opened 2020-04-14
Add generic libvirt connection plugin

odyssey4me:
##### SUMMARY

Initial pass at libvirt/qemu connection plugin

Fixes: #12 

##### ISSUE TYPE

- Feature Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/connection/libvirt_qemu.py
+++ b/plugins/connection/libvirt_qemu.py
@@ -1,0 +1,279 @@
+# Based on local.py (c) 2012, Michael DeHaan <michael.dehaan@gmail.com>
+# Based on chroot.py (c) 2013, Maykel Moya <mmoya@speedyrails.com>
+# (c) 2013, Michael Scherer <misc@zarb.org>
+# (c) 2015, Toshio Kuratomi <tkuratomi@ansible.com>
+# (c) 2017 Ansible Project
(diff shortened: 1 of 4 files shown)


daveol:
Should this not be named libvirt_qemu? Since it is only compatble with the qemu guest agent at the moment...

odyssey4me:
@daveol I can do that. I think that's sensible. I'm hoping to extend its functionality, but time is a bit tight - so maybe it'd be better to rename it for now and once we have something else we can shift it over and leave an alias.

odyssey4me:
@daveol Should I merge it as-is, then work on the additional remote URI functionality in a subsequent PR? What do you think? I'm thinking this may be useful as a base and I'd like to unblock any work you'd like to do on top of it.

daveol:
@odyssey4me you can merge as-is if you want, the ```libvirt_lxc``` connection plugin would probably also need an update for supporting the remote URI functionality to keep it consistent (also an update to use the libvirt python api, but that's a different issue).

Then again it does not seem like a lot of work to implement the URI functionality for this plugin,  or am I wrong?

odyssey4me:
@daveol I think I agree. While it's not a lot of work, I don't have enough time right now to dig into the details of it. So meh, I'll register an issue for it and we can move along. :)

PR #30, opened 2020-04-18
Fix doc issues flagged by `ansible-test sanity`

daveol:
##### SUMMARY
This fixes some missing types, aliases and choices.

##### ISSUE TYPE
- Docs Pull Request

### DIFF
--- a/plugins/modules/virt.py
+++ b/plugins/modules/virt.py
@@ -25,6 +25,9 @@ options:
       - name of the guest VM being managed. Note that VM must be previously
         defined with xml.
       - This option is required unless I(command) is C(list_vms) or C(info).
+    type: str
+    aliases:
+      - guest
   state:
     description:
       - Note that there may be some lag for state requests like C(shutdown)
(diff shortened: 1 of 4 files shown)


plugins/modules/virt.py
```diff
@@ -575,7 +582,7 @@ def main():
     module = AnsibleModule(
         argument_spec=dict(
             name=dict(type='str', aliases=['guest']),
-            state=dict(type='str', choices=['destroyed', 'pause', 'running', 'shutdown']),
+            state=dict(type='str', choices=['destroyed', 'paused', 'running', 'shutdown']),
```
odyssey4me, line 585: According to https://docs.ansible.com/ansible/latest/modules/virt_module.html it's 'pause'. If we switch to 'paused' then we'll need an alias for backwards compatibility.

daveol: I think you confused it with command, which has a 'pause' command... this is state, where the documentation was correct but the argument_spec was incorrect

PR #45, opened 2020-09-01
[WIP] Add support for managing libvirt volumes

galli-leo:
##### SUMMARY
This PR adds support for defining libvirt volumes (via `libvirt_volume`).
Currently, this is in a very basic WIP state, but I was able to define a volume and list existing volumes :)

Opening it up, since I have some questions regarding how to implement certain stuff best (see below).

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
libvirt_volume

##### ADDITIONAL INFORMATION
To start out, I just copied the stuff from `libvirt_pool` over and modified it to my needs.
I also need to query pool information, so I refactored the `Libvirtconnection` out of `libvirt_pool` into a `module_utils`. It should probably be renamed now.
However, I am not entirely sure whether this is the intended way to do this inside an ansible collection? I was only able to import it from there by supplying the fully qualified path, which seems a bit silly.

Furthermore, how should the idempotency be handled here? For example, if I have the following in a playbook:

```yaml
    - name: Create Volume
      community.libvirt.virt_volume:
        name: main_os
        pool: vg_vm_storage
        state: present
        xml: '{{ lookup("file", "./templates/os_disk.xml") }}'
```

And I change the template to have less / more storage, should the volume be resized automatically?

### DIFF
--- a/plugins/module_utils/entry.py
+++ b/plugins/module_utils/entry.py
@@ -1,0 +1,2 @@
+class EntryNotFound(Exception):
+    pass
--- a/plugins/module_utils/pool.py
+++ b/plugins/module_utils/pool.py
@@ -1,0 +1,220 @@
+try:
+    import libvirt
+except ImportError:
+    HAS_VIRT = False
+else:
+    HAS_VIRT = True
+
+from ansible_collections.community.libvirt.plugins.module_utils.entry import EntryNotFound
+
+ENTRY_STATE_ACTIVE_MAP = {
+    0: "inactive",
+    1: "active"
+}
+
+ENTRY_STATE_AUTOSTART_MAP = {
+    0: "no",
+    1: "yes"
+}
+
+ENTRY_STATE_PERSISTENT_MAP = {
+    0: "no",
+    1: "yes"
+}
+
+class LibvirtConnection(object):
+
+    def __init__(self, uri, module):
+
+        self.module = module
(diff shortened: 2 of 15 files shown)


odyssey4me:
This is nice work, thanks for doing it! Thanks especially for the refactoring to a common module_utils. I've been meaning to do that for some time.

To answer your questions:

1. Yes, the way to import module_utils is to use the full collections path as far as I can see by inspecting other collections.
2. With regards to the idempotancy question - if the volume supports the resize, then it seems sensible to me to do it. If it does not then the task should fail with an appropriate error to indicate resizing is not supported if that makes sense and is possible. Perhaps supporting resizing could be a module argument which is enabled by default, but can be disabled.

odyssey4me:
Could you rebase this on to the head of main please?

Andersson007:
Hi @galli-leo , thanks for the PR!
Could you rebase this on to the head of main please https://docs.ansible.com/ansible/latest/dev_guide/developing_rebasing.html ?
If you are not interested in continuing working on this, please put it explicitly. We could close the PR and create an issue linking to the PR. Maybe anyone could pick it up and finish later.

Andersson007:
As there's been no response, I'll close the PR. If you get interested in continuing this, we'll reopen it. Thanks!

Issue #99, opened 2021-12-25
inventory plugin: add a way to skip dormant domains


akdev1l:
##### SUMMARY

Currently I have some libvirt domains that I created that are meant to be shutdown most of the time, this means doing something like: `ansible -m ping all` will always return errors like the following:

```
libvirt: Domain Config error : Requested operation is not valid: domain is not running
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: libvirt.libvirtError: Requested operation is not valid: domain is not running
nafta.hq.akdev.xyz | FAILED! => {
    "msg": "Unexpected failure during module execution.",
    "stdout": ""
}
```

This clutters the output and it probably makes playbooks take slightly longer as it tries and fails each command for domains that aren't running.

My proposal would be to add a new option in the config file:

```
plugin: community.libvirt.libvirt
uri: qemu:///system
ignore_off: True
```

with such configuration the plugin would just return a list of domains that are currently active. (it could warn that other domains are being ignored for awareness)

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

community.libvirt.libvirt inventory plugin

##### ADDITIONAL INFORMATION

If the feature is implemented then users would be able to run playbooks without getting errors whenever they have one or more VMs that are shut down.

```
$ ansible -m ping all
libvirt: Domain Config error : Requested operation is not valid: domain is not running
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: libvirt.libvirtError: Requested operation is not valid: domain is not running
nafta.hq.akdev.xyz | FAILED! => {
    "msg": "Unexpected failure during module execution.",
    "stdout": ""
}
```

---

I am willing to put some time towards making this happen if no one is against having this - alternatively if there's a way of achieving what I want that I'm unaware of please let me know.

csmart:
Hi @akdev1l , thanks for suggesting this idea. It's an interesting one which I'll need to think about and discuss some more, I think. My first instinct is to suggest that the behaviour of the ping module when using the dynamic inventory should be the same as other inventories. In my experience, it's normal for ping to fail against a host if it is not contactable. For example, if this was a dynamic inventory from OpenStack or something which returned a list of hosts, if some of those were off it would try and then also fail, as you're seeing. The failure also means that those hosts are then taken out of the in-memory inventory and excluded from further tasks.

[...]

PR #101, opened 2022-01-06
Make the connection plugin optional and add the option to filter domains / uuids using regex.

yorick1989:
…ins / uuids using regex.

##### SUMMARY
I ran into issues that the connection plugin was not working for out-of-the-box Red Hat servers. Red Hat has disabled the required Qemu Guest Agent commands for the libvirt connection plugin by default:
```bash
[root@test01 ~]# cat /etc/redhat-release 
Red Hat Enterprise Linux release 8.5 (Ootpa)
[root@test01 ~]# dnf download --downloadonly --destdir=/tmp qemu-guest-agent
Updating Subscription Management repositories.
Last metadata expiration check: 2:05:11 ago on Thu 06 Jan 2022 07:05:05 PM CET.
qemu-guest-agent-4.2.0-59.module+el8.5.0+13495+8166cdf8.1.x86_64.rpm                                                                    601 kB/s | 257 kB     00:00    
[root@test01 ~]# rpm2cpio /tmp/qemu-guest-agent-4.2.0-59.module+el8.5.0+13495+8166cdf8.1.x86_64.rpm | cpio -i --to-stdout ./etc/sysconfig/qemu-ga | grep BLACKLIST
BLACKLIST_RPC=guest-file-open,guest-file-close,guest-file-read,guest-file-write,guest-file-seek,guest-file-flush,guest-exec,guest-exec-status
741 blocks
```

Therefor I require to pull the IP addresses for each domain and set the first IP address found as the `ansible_host` variable.

I also added the `ansible_libvirt_ifaces` variable which contains all the network interfaces information of the domain. It's not necessary, but I was already working on it to get `ansible_host` to work anyway :) .

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
no_connection_plugin

##### ADDITIONAL INFORMATION

*Connection plugin:*

Before:
```bash
$ ansible-inventory --host test01
{
    "ansible_connection": "community.libvirt.libvirt_qemu",
    "ansible_libvirt_uri": "qemu:///system"
}
```

After (set: `use_connection_plugin: False`):
```bash
$ ansible-inventory --host test01
{
    "ansible_host": "192.168.122.182",
    "ansible_libvirt_ifaces": {
        "vnet39": {
            "addrs": [
                {
                    "addr": "192.168.122.182",
                    "prefix": 24,
                    "type": 0
                }
            ],
            "hwaddr": "52:54:00:12:54:a2"
        }
    }
}
```

*Filter*

Before:
```bash
$ ansible-inventory --graph | grep -v -- '-@'
@all:
  |  |--test01
  |  |--test03
  |  |--test01
  |  |--test01
  |  |--test03
  |  |--kali
  |  |--test02
  |  |--test03
```

After (set: `filter: "test0(1|3)+"`):
```bash
$ ansible-inventory --graph | grep -v -- '-@' | sort -n | uniq
@all:
  |  |--test01
  |  |--test03
  |  |--test01
  |  |--test01
  |  |--test03
  |  |--test03
```

### DIFF
--- a/plugins/inventory/libvirt.py
+++ b/plugins/inventory/libvirt.py
@@ -20,7 +20,7 @@ options:
     uri:
         description: Libvirt Connection URI
         required: True
-        type: str
+        type: string
     inventory_hostname:
         description: |
             What to register as the inventory hostname.
@@ -35,6 +35,16 @@ options:
             - name
             - uuid
         default: "name"
+    use_connection_plugin:
+        description: Whether or not to use the connection plugin.
+        type: boolean
+        default: True
+    filter:
+        description: |
+            Use a regex string filter out specific domains (by name or uuid;
+            this depends on inventory_hostname).
+        type: string
+        default: ".*"
 requirements:
     - "libvirt-python"
 '''
@@ -48,6 +58,8 @@ uri: 'lxc:///'
 plugin: community.libvirt.libvirt
 uri: 'qemu:///system'
 '''
+
+import re
 
 from ansible.plugins.inventory import BaseInventoryPlugin, Constructable
 from ansible.errors import AnsibleError
@@ -92,12 +104,26 @@ class InventoryModule(BaseInventoryPlugin, Constructable):
 
         # TODO(daveol)
         # make using connection plugins optional
-        connection_plugin = dict({
-            'LXC': 'community.libvirt.libvirt_lxc',
-            'QEMU': 'community.libvirt.libvirt_qemu'
-        }).get(connection.getType())
+        use_connection_plugin = self.get_option('use_connection_plugin')
+
+        if use_connection_plugin:
+            connection_plugin = dict({
+                'LXC': 'community.libvirt.libvirt_lxc',
+                'QEMU': 'community.libvirt.libvirt_qemu'
+            }).get(connection.getType())
+
+        # Set the domain filter.
+        _filter = self.get_option('filter')
 
         for server in connection.listAllDomains():
+            if not dict({
+                   'uuid': re.match(_filter, server.UUIDString()),
+                   'name': re.match(_filter, server.name())
+               }).get(
+                   self.get_option('inventory_hostname')
+               ):
+                continue
+
             inventory_hostname = dict({
                 'uuid': server.UUIDString(),
                 'name': server.name()
@@ -118,7 +144,32 @@ class InventoryModule(BaseInventoryPlugin, Constructable):
             self.inventory.add_group(inventory_hostname_alias)
             self.inventory.add_child(inventory_hostname_alias, inventory_hostname)
 
-            if connection_plugin is not None:
+            # Set the interface information.
+            ifaces = {}
+
+            for iface, iface_info in (connection.lookupByName(server.name())).interfaceAddresses(
+                                      libvirt.VIR_DOMAIN_INTERFACE_ADDRESSES_SRC_LEASE).items():
+                # Set interface hw address.
+                ifaces[iface] = {
+                  'hwaddr': iface_info['hwaddr'],
+                  'addrs': []
+                }
+
+                if 'addrs' in iface_info:
+
+                    # Append the addresses information to the interface dict.
+                    ifaces[iface]['addrs'] = iface_info['addrs']
+
+                    # Set the ansible_host variable to the first IP address found.
+                    if not use_connection_plugin and len(iface_info['addrs']) >= 1 and \
+                       'ansible_host' not in self.inventory.hosts[inventory_hostname].get_vars():
+                        self.inventory.set_variable(
+                            inventory_hostname,
+                            'ansible_host',
+                            iface_info['addrs'][0]['addr']
+                        )
+
+            if use_connection_plugin and connection_plugin is not None:
                 self.inventory.set_variable(
                     inventory_hostname,
                     'ansible_libvirt_uri',
@@ -129,6 +180,12 @@ class InventoryModule(BaseInventoryPlugin, Constructable):
                     'ansible_connection',
                     connection_plugin
                 )
+
+            self.inventory.set_variable(
+                inventory_hostname,
+                'ansible_libvirt_ifaces',
+                ifaces
+            )
 
             # Get variables for compose
             variables = self.inventory.hosts[inventory_hostname].get_vars()


odyssey4me:
It seems to me that there is already functionality to not use the connection plugin, by simply removing the connection plugin configuration file?

Also, the resolving of libvirt domains to IP addresses is the sort of thing that belongs in an inventory plugin, surely?

If you're wanting to connect via SSH then the inventory plugin should resolve libvirt domains to IP addresses, then the normal SSH connection plugin does the usual from there.

yorick1989:
Thank you for your feedback.

How do I remove the 'connection plugin configuration file'? I only use the inventory filename that contains:
```yaml
plugin: community.libvirt.libvirt
uri: 'qemu:///system'
use_connection_plugin: False
filter: "test0(1|3)+"
```

By the way, my changes are done in the inventory code, not in the connection plugin code. There was no option to skip the connection plugin within the libvirt inventory file, that is why I added the option to skip the connection plugin by adding the `use_connection_plugin` option to use an alternative way which pulls the IP address and adds the first found IP address to the host variables (as `ansible_host`); so this IP address can used for SSH connections. [There was also a comment in the inventory code that making the connection plugin optional](https://github.com/ansible-collections/community.libvirt/pull/101/commits/d62178b2c21c4b1d5af39a117b8443363b4cfefd#diff-462ec562a7affc2f50dc4168b947e4d666b1a2af559e431bad0a715d4d941f73L93-L98). This is why I made these changes and I would like to give it back to the upstream; if it is a beneficial addition.

If there is a better alternative, please let me know. I just try a way to get this to work without using the connection plugin; since that's by default not possible on Red Hat servers (since the 'guest-file-open,guest-file-close,guest-file-read,guest-file-write,guest-file-seek,guest-file-flush,guest-exec,guest-exec-status' options are blocked for the `qemu-guest-agent`).

Thanks!

csmart:
Hi @yorick1989 thanks for this contribution, I will test it out. i can see that on RHEL machines qemu guest agent does not support the required options so I think another option is useful :+1:  I may have confused @odyssey4me when I was talking to him about this, as I incorrectly said it was in the connection not the inventory code. Sorry!

If we are going to rely on getting the IP from libvirt then I feel like it has to be very robust. Also, how will we handle failure scenarios? I'm wondering how this might work in edge cases, such as when an IP is not available yet, or perhaps only an IPv6 (still waiting for IPv4), or guests which have multiple network interfaces, etc. Have you had a chance to test any of those scenarios?

Also, are there any additional requirements for this to work? Thanks!

csmart:
I've done some basic testing with an inventory file like so:

```yaml
---
plugin: community.libvirt.libvirt
uri: 'qemu:///system'
use_connection_plugin: False
filter: "ansible-fedora-35"
```

it works as expected, SSHing into the guest:

```bash
(3.9) [09:40 csmart@dev ~/.../community/libvirt (connection_filter $%)]$ date
Tue 01 Feb 2022 09:40:05 AEDT

(3.9) [09:40 csmart@dev ~/.../community/libvirt (connection_filter $%)]$ ansible -i ../../kvm.yml all --limit ansible-fedora-35 -m command -a 'whoami'
ansible-fedora-35 | CHANGED | rc=0 >>
csmart

(3.9) [09:40 csmart@dev ~/.../community/libvirt (connection_filter $%)]$ ssh ansible-fedora-35 last |head -1
csmart   pts/0        192.168.112.1    Tue Feb  1 09:40 - 09:40  (00:00)
```

I have tested with a `filter` applied but `use_connection_plugin` set to `false` so that it uses guest agent rather than SSH:

```yaml
---
plugin: community.libvirt.libvirt
uri: 'qemu:///system'
#use_connection_plugin: False
filter: "ansible-fedora-35"
```

It works as expected.

```bash
(3.9) [09:45 csmart@dev ~/.../community/libvirt (connection_filter $%)]$ ansible -i ../../kvm.yml all -m command -a 'whoami'
ansible-fedora-35 | CHANGED | rc=0 >>
root
```

I have tested using SSH but when there is no `filter` specified in the inventory:

```yaml
---
plugin: community.libvirt.libvirt
uri: 'qemu:///system'
use_connection_plugin: False
#filter: "ansible-fedora-35"
```

However Ansible is not able to parse the inventory. I think we would need to make sure that by default, an empty or missing filter resulted in a complete list of all VMs (the default without this code).

```bash
(3.9) [09:42 csmart@dev ~/.../community/libvirt (connection_filter $%)]$ ansible -i ../../kvm.yml all -m command -a 'whoami'
libvirt: Domain Config error : Requested operation is not valid: domain is not running
[WARNING]:  * Failed to parse /home/csmart/ansible_collections/kvm.yml with
ansible_collections.community.libvirt.plugins.inventory.libvirt plugin: Requested operation is not valid: domain is not running
[WARNING]: Unable to parse /home/csmart/ansible_collections/kvm.yml as an inventory source
[WARNING]: No inventory was parsed, only implicit localhost is available
[WARNING]: provided hosts list is empty, only localhost is available. Note that the implicit localhost does not match 'all'
```

I've also tested with `use_connection_plugin: True` and it works as expected.

Cheers!

Andersson007:
@csmart thanks for testing and the great report! @yorick1989 @odyssey4me any thoughts / related updates?

plugins/inventory/libvirt.py
```diff
@@ -118,7 +144,32 @@ def parse(self, inventory, loader, path, cache=True):
             self.inventory.add_group(inventory_hostname_alias)
             self.inventory.add_child(inventory_hostname_alias, inventory_hostname)
 
-            if connection_plugin is not None:
+            # Set the interface information.
+            ifaces = {}
+
+            for iface, iface_info in (connection.lookupByName(server.name())).interfaceAddresses(
```
odyssey4me, line 150: This is a lot of work to go through to throw away if the connection plugin is going to be used. Are we doing this so that we add the interface information to the inventory, whether we use the connection plugin or not?

yorick1989: Yes, it was my intention to have the interface information available in both cases: connection plugin on or off.I can leave it out in case that's a bit overhead. But I can imagine that in some cases you would like to know this kind of information (if someone would like to make modifications within libvirt itself for example).

Another option is to leave it out when the connection plugin is on and add it later on back, together with all the other information that's available from libvirt about a host; so you have a full set of information about each host from libvirt. For example: CPU info, RAM info, DISK info, etc.

I'm fine with any of these options. Or if you have a better one, please let me know!

yorick1989:
Thank you all for the feedback!

@csmart Thank you for testing! I also tested it myself and it works for me:
```bash
[yorick@yorick-pc ansible]$ cat libvirt_qemu.yml 
plugin: community.libvirt.libvirt
uri: 'qemu:///system'
use_connection_plugin: False
#filter: "test-4[0-9]+"
[yorick@yorick-pc ansible]$ ansible -i libvirt_qemu.yml all -m ping
[WARNING]: Invalid characters were found in group names but not replaced, use -vvvv to see details
kali | UNREACHABLE! => {
    "changed": false,
    "msg": "Failed to connect to the host via ssh: ansible@192.168.122.129: Permission denied (publickey,password).",
    "unreachable": true
}
test04 | SUCCESS => {
    "ansible_facts": {
        "discovered_interpreter_python": "/usr/libexec/platform-python"
    },
    "changed": false,
    "ping": "pong"
}
test01 | SUCCESS => {
    "ansible_facts": {
        "discovered_interpreter_python": "/usr/libexec/platform-python"
    },
    "changed": false,
    "ping": "pong"
}
test03 | SUCCESS => {
    "ansible_facts": {
        "discovered_interpreter_python": "/usr/libexec/platform-python"
    },
    "changed": false,
    "ping": "pong"
}
```

So, it's not clear to me why it failed on your side. The filter uses regex and by default the filter value is set to `.*`; so that should match every host by default.

Also thanks for the feedback in your first reply. Here my replies to your questions:
- How will we handle failure scenarios? \
  _Do you mean that I have to add more debug and error handling to my code?_
- I'm wondering how this might work in edge cases, such as when an IP is not available yet, or perhaps only an IPv6 (still waiting for IPv4), or guests which have multiple network interfaces, etc. \
  _This was indeed also a struggle point for me: decide which IP can be reached over SSH if there are multiple ones. I don't know how to add dynamic filters to the inventory file / plugin script to tell which interface or IP-address should be used for the SSH-connection; since network interfaces can differ for each of the available servers. That's why I also added the `ansible_libvirt_ifaces` dict which can be used to filter out the proper IP-addresses and set that proper one as `ansible_host` yourself in a (pre-executed) playbook. I know it's ugly, but if one of you can give me a better solution, please feel free to advice me! \
  And I don't know how to act on servers which don't have an IP-address available (,yet). If the inventory plugin simply can't find it in libvirt, then the logic within the inventory plugin should set the `ansible_libvirt_ifaces` variable to an empty list. Or is that too simple? =) \
  And IPv6 addresses shouldn't be a problem I guess, if the host can be reached on it. Or am I missing something?_
- How will it work with multiple inventories, or where one has specified ansible_host already elsewhere? \
  _I can add a check for `ansible_host`, before setting it. But shouldn't the last sourced inventory overwrite the already existing variables by default?_

As you can see, I'm pretty new with this kind of modifications. I needed this modification to deploy / destroy my servers  on a quickly manner using terraform. So any help to improve my changes is very welcome!

csmart:
There's [an issue](https://github.com/ansible-collections/community.libvirt/issues/99) with a request for filtering by VM state, for example, power off. I think if we're adding filtering here, it would be great to have this in mind so that we can filter by name (current) but also work in states at some point.

yorick1989:
I've seen the activity on this old pull request. It has been a while, but if you want me to help with anything, please let me know.

csmart:
@thinkdoggie I've created new MR #257 to replace #101, keeping @yorick1989 as the author, that is rebased on main and keeps only the filter change, as per your review. I'm going to close this MR and if we still want this feature, then we can merge #257 instead (or close it too, if we don't want it). Thanks again @yorick1989 for this idea, feel free to comment on #257 if you like!

PR #106, opened 2022-02-02
CI: remove centos8 from test matrix

csmart:
##### SUMMARY

CentOS 8 has been fully discontinued and mirrors are removing content.

This patch removes centos8 from the test matrix.

Relates to https://github.com/ansible-collections/news-for-maintainers/issues/3

##### ISSUE TYPE

- CI Pull Request

### DIFF
--- a/.azure-pipelines/azure-pipelines.yml
+++ b/.azure-pipelines/azure-pipelines.yml
@@ -135,8 +135,6 @@ stages:
           targets:
             - name: CentOS 7
               test: centos7
-            - name: CentOS 8
-              test: centos8
             - name: Fedora 33
               test: fedora33
             - name: Fedora 34
@@ -162,8 +160,6 @@ stages:
               test: centos6
             - name: CentOS 7
               test: centos7
-            - name: CentOS 8
-              test: centos8
             - name: Fedora 32
               test: fedora32
             - name: Fedora 33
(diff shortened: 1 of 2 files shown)


csmart:
I'm not sure whether replacing `centos8` with `rhel8.5` will actually work... so let's see. If its' not possible, we can just remove `centos8` entirely.

Andersson007:
I think RHEL should go to Remote section, there are currently a couple of versions

Andersson007:
test https://github.com/ansible-collections/news-for-maintainers/issues/3

csmart:
> I'm not sure whether replacing `centos8` with `rhel8.5` will actually work... so let's see. If it's not possible, we can just remove `centos8` entirely. I'm also not sure whether shippable should also be changed, but have done so pending review.

So no... rhel8.5 only works with certain Ansible versions as defined in Ansible core. So this patch has been updated to just remove centos8 and leave it at that.

In another commit we might clean up shippable.yml as it is not needed, but I'd rather not mix that in with this centos8 specific change.

lestercheung:
Noob question: do we test for CentOS Stream 8 at all?

Andersson007 (approved):
LGTM

Andersson007:
@lestercheung I don't think so (feel free to ask any questions that come to mind; there are also active [real time chats](https://docs.ansible.com/ansible/latest/community/communication.html#real-time-chat) available for communication)

lestercheung:
I do join #user from time to time but will start joining #devel and learn a thing or two. 😉

PR #114, opened 2022-04-18
Remove usage of distutils.spawn

csmart:
distutils has been deprecated and will be removed from Python's stdlib
in Python 3.12 (see https://www.python.org/dev/peps/pep-0632/). This PR
replaces the use of deprecated ``distutils.spawn.find_executable`` with
Ansible's ``get_bin_path`` in ``_search_executable`` function.

See https://github.com/ansible-collections/overview/issues/45#issuecomment-999911221


csmart:
The tests are failing due to updated documentation tests, therefore this PR is abandoned and the change bundled with #115 instead.

PR #135, opened 2022-08-11
Replace discouraged function listVolumes

bturmann:
This PR is similar to #134.

##### SUMMARY
According to the libvirt storage API reference, the function

- listVolumes (virStoragePoolListVolumes)

should not be used anymore: "The use of this
function is discouraged. Instead, use virStoragePoolListAllVolumes()."

Source:
https://libvirt.org/html/libvirt-libvirt-storage.html#virStoragePoolListVolumes

#### Compatibility / Risk:
The function "virStoragePoolListAllVolumes()" appeared in version
0.10.2 which was released in 2012. It seems rather unlikely that
somebody is still using an older unsupported libvirt version, so the
risk should be rather low.

Source:
https://libvirt.org/hvsupport.html#virStorageDriver

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
libvirt_pool

##### ADDITIONAL INFORMATION
My tests confirm that the bugfix works as expected. It does not change the output of virt_pool with facts and info compared to the old function.

### DIFF
--- a/changelogs/fragments/135_virt_pool_replace_function_listVolumes.yml
+++ b/changelogs/fragments/135_virt_pool_replace_function_listVolumes.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - virt_pool - replace discouraged function ``listAllVolumes`` with ``listAllVolumes`` to fix potential race conditions (https://github.com/ansible-collections/community.libvirt/pull/135).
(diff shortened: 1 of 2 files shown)


csmart:
This looks good, thanks @bturmann.

Andersson007:
@bturmann @csmart thanks!

Issue #81, opened 2021-07-21
Release plan


Andersson007:
##### SUMMARY
_(partially copied from https://github.com/ansible-collections/community.crypto/issues/74)_

We should decide eventually on how to release this collection (w.r.t. versioning).
Small collections like this one don't need a complex plan like the one for community.general and community.network.
So how about the following?

1. Release minor and patch releases whenever we want (like after adding new features or fixing bugs). Since this collection is small, there's no need to fix things in advance. Just add features, and after a feature either wait a bit longer for more features/bugs, or make a release.

I suggest releasing without branching https://github.com/ansible/community/wiki/ReleasingCollections#releasing-without-release-branches-for-smaller-collections

Once we release a 2.0.0 (with some breaking change relative to 1.x.y), we can have a stable-1 branch so we can backport bugfixes (or even features) if needed, and release more 1.x.y versions.

Andersson007:
I'm gonna release version 1.0.2 tomorrow (22.07.2021).

Andersson007:
community.libvirt 1.0.2 has just been released

thinkl33t:
Can we push 1.1.1 (or 1.2.0) with the changes from this weekend? :)

Andersson007:
There's also a [minor change](https://github.com/ansible-collections/community.libvirt/tree/main/changelogs/fragments) made since the last release, so it should be 1.2.0:)

bturmann:
Hi,
can we release 1.2.x?

csmart:
Oh yes, thanks for the bump. I will arrange a release once !157 is merged, which I expect will be this week.

csmart:
Pipeline issues have been merged in #158 and #157 has been merged, so we are unblocked now to do a release, which is the next priority.

csmart:
`1.3.0` has been tagged and is now available on Galaxy, thank you to everyone for your contributions towards this release! Please let me know if you have any issues.

bturmann:
@csmart, many thanks for the new release 1.3.0.

leegarrett:
Time for a new release? I've seen that #147, #156, and #162 have been merged, and I'd prefer to drop them from the Debian package of ansible-community by pulling an upstream update.

csmart:
Hi @leegarrett , happy to look at another release if needed. I actually haven't merged #162 yet, see my comments on that MR. I think the others should be available already in v1.3.0? What version are you paying for Debian?

thinkdoggie:
[...]

PR #136, opened 2022-08-12
Fix virt module to undefine a domain with nvram or other metadata

### LINKED ISSUE #40: virt module is not able to undefine a domain with defined nvram file
##### SUMMARY

Virt module is not able to undefine a domain with defined nvram file.
migrated from: https://github.com/ansible/ansible/issues/43915
##### ISSUE TYPE

 - Bug Report

##### COMPONENT NAME

virt

##### ANSIBLE VERSION

```
ansible 2.6.2
python version = 2.7.12
```

##### STEPS TO REPRODUCE

```yaml
- name: undefine vm with nvram
  virt:
    name: example-vm-nvram
    command: undefine
```

##### EXPECTED RESULTS

domain "example-vm-nvram" is undefined
##### ACTUAL RESULTS

```
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: libvirt.libvirtError: Requested operation is not valid: cannot delete inactive domain with nvram
fatal: [localhost]: FAILED! => {"changed": false, "failed_when_result": true, "msg": "Requested operation is not valid: cannot delete inactive domain with nvram"}
```
The libvirt function ```undefine``` is not able to remove a domain with nvram, by using ```undefineFlags``` it is possible to control the deletion of the nvram file. See:
https://libvirt.org/html/libvirt-libvirt-domain.html#virDomainUndefine
https://libvirt.org/html/libvirt-libvirt-domain.html#virDomainUndefineFlags

With the following command the undefine action is possible.
```
virsh undefine example-vm-nvram --nvram
```

bturmann:
##### SUMMARY
Fixes #40

Libvirt function `undefine()` is not able to delete nvram or other
metadata. Therefore it is replaced with `undefineFlags()` which is able to
handle it by using flags.

All possible flags are listed in 'ENTRY_UNDEFINE_FLAGS_MAP'. Integer 23
makes `undefine` successful in all cases (1 + 2 + 4 + 16).

Source:
https://libvirt.org/html/libvirt-libvirt-domain.html#virDomainUndefineFlags

The flags `nvram`(4) and `keep_nvram`(8) are mutually exclusive.

To control the metadata handling during `undefine`, two new options
`force` and `flags` are introduced including documentation and examples.

#### Compatibility / Risk:
The function `undefineFlags()` appeared in libvirt version 0.9.4 which was
release in 2011. It seems rather unlikely that somebody is still using
an older unsupported libvirt version. Additionally, if none of the new
module options are provided, then it behaves as before and maintains
backward compatibility, means the overall risk of this commit should be
rather low.

Source:
https://libvirt.org/hvsupport.html#virHypervisorDriver

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
virt

##### ADDITIONAL INFORMATION
My tests confirm that the bugfix works as expected, like `undefine` a qemu guest with UEFI nvram or a qemu guest with BIOS. Also other commands are still working, like `start`, `shutdown` or `destroy`.

ansible adhoc command:
```paste below
$ ansible -b -m community.libvirt.virt -a "name=guest99 command=undefine force=true" host99
```

### DIFF
--- a/plugins/modules/virt.py
+++ b/plugins/modules/virt.py
@@ -16,6 +16,29 @@ module: virt
 short_description: Manages virtual machines supported by libvirt
 description:
      - Manages virtual machines supported by I(libvirt).
+options:
+    flags:
+        choices: [ 'managed_save', 'snapshots_metadata', 'nvram', 'keep_nvram', 'checkpoints_metadata']
+        description:
+            - Pass additional parameters.
+            - Currently only implemented with command C(undefine).
+              Specify which metadata should be removed with C(undefine).
+              Useful option to be able to C(undefine) guests with UEFI nvram.
+              C(nvram) and C(keep_nvram) are conflicting and mutually exclusive.
+              Consider option C(force) if all related metadata should be removed.
+        type: list
+        elements: str
+    force:
+        description:
+            - Enforce an action.
+            - Currently only implemented with command C(undefine).
+              This option can be used instead of providing all C(flags).
+              If C(yes), C(undefine) removes also any related nvram or other metadata, if existing.
+              If C(no) or not set, C(undefine) executes only if there is no nvram or other metadata existing.
+              Otherwise the task fails and the guest is kept defined without change.
+              C(yes) and option C(flags) should not be provided together. In this case
+              C(undefine) ignores C(yes), considers only C(flags) and issues a warning.
+        type: bool
 extends_documentation_fragment:
     - community.libvirt.virt.options_uri
     - community.libvirt.virt.options_xml
@@ -66,6 +89,30 @@ EXAMPLES = '''
     command: define
     xml: "{{ lookup('template', 'vm_template.xml.j2') }}"
     autostart: yes
+
+# Undefine VM only, if it has no existing nvram or other metadata
+- name: Undefine qemu VM
+  community.libvirt.virt:
+    name: foo
+
+# Undefine VM and force remove all of its related metadata (nvram, snapshots, etc.)
+- name: "Undefine qemu VM with force"
+  community.libvirt.virt:
+    name: foo
+    force: yes
+
+# Undefine VM and remove all of its specified metadata specified
+# Result would the same as with force=true
+- name: Undefine qemu VM with list of flags
+  community.libvirt.virt:
+    name: foo
+    flags: managed_save, snapshots_metadata, nvram, checkpoints_metadata
+
+# Undefine VM, but keep its nvram
+- name: Undefine qemu VM and keep its nvram
+  community.libvirt.virt:
+    name: foo
+    flags: keep_nvram
 
 # Listing VMs
 - name: List all VMs
@@ -134,6 +181,17 @@ VIRT_STATE_NAME_MAP = {
     6: 'crashed',
 }
 
+ENTRY_UNDEFINE_FLAGS_MAP = {
+    'managed_save': 1,
+    'snapshots_metadata': 2,
+    'nvram': 4,
+    'keep_nvram': 8,
+    'checkpoints_metadata': 16,
+}
+
+ALL_FLAGS = []
+ALL_FLAGS.extend(ENTRY_UNDEFINE_FLAGS_MAP.keys())
+
 
 class VMNotFound(Exception):
     pass
@@ -198,8 +256,8 @@ class LibvirtConnection(object):
     def destroy(self, vmid):
         return self.find_vm(vmid).destroy()
 
-    def undefine(self, vmid):
-        return self.find_vm(vmid).undefine()
+    def undefine(self, vmid, flag):
+        return self.find_vm(vmid).undefineFlags(flag)
 
     def get_status2(self, vm):
         state = vm.info()[0]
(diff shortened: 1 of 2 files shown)


csmart:
This looks good, thanks @bturmann, but I need to go over it a bit carefully as I think this will change the behaviour quite a bit that people may have been relying on - for example, users may rely on the fact that nvram isn't cleaned up, or that VMs with snapshots aren't removed.

Andersson007:
if it's a behavioral change, it must be announced now (with `major_changes:`) but released in the next major release (the PR must be merged right before it happens).

bturmann:
Agreed. What do you think about following?

- introduce a new parameter "force", similar to what other modules do
- only If force would be set to "True" a guest would be undefined even though having nvram or other metadata
- if force is not defined (should be default) or set to false, the behaviour is as of today being fully backward compatible
- explain parameter force in module doc

Andersson007:
> Agreed. What do you think about following?
> 
> * introduce a new parameter "force", similar to what other modules do
> * only If force would be set to "True" a guest would be undefined even though having nvram or other metadata
> * if force is not defined (should be default) or set to false, the behaviour is as of today being fully backward compatible
> * explain parameter force in module doc

Sounds like a solution to me, @csmart what do you think?

csmart:
That sounds excellent! Thanks.

plugins/modules/virt.py
```diff
@@ -539,7 +543,10 @@ def core(module):
                 # Provide flag as an integer of all desired bits, see 'ENTRY_UNDEFINE_FLAGS_MAP'.
                 # Integer 23 takes care of all cases (23 = 1 + 2 + 4 + 16).
                 flag = 0
-                if force is True:
+                if mode is not None:
+                    for m in mode:
+                        flag += ENTRY_UNDEFINE_FLAGS_MAP.get(m, 0)
```
csmart, line 548: @bturmann both `nvram` and `keep_nvram` are conflicting options, aren't that? I'm not sure what happens if both are set, but maybe we should have a check for this?

bturmann: Very good catch, @csmart. Indeed, they are conflicting options. I omitted a check, because my tests confirmed that libvirt library already refuses it with following rather clear error message:

```
$ ansible -b -m community.libvirt.virt -a "name=guest99 command=undefine mode=nvram,keep_nvram" host99
An exception occurred during task execution. To see the full traceback, use -vvv.
The error was: libvirt.libvirtError: 
Requested operation is not valid: cannot both keep and delete nvram
host99 | FAILED! => {
    "changed": false,
    "msg": "Requested operation is not valid: cannot both keep and delete nvram"
}
```

Therefore, another check here seems unnecessary. What do you think?

csmart: Nice! I guess given that we're making it more advanced (and safe by default), we could leave it there and let it fail if someone happens to use conflicting flags. We just need to make sure we pass the error up the user, I guess.

Although my other thought is around what happens if the upstream behaviour changes, or different distros do something different? So perhaps an alternative is to implement a check ourselves and fail (protecting against upstream/distro changes and making the behaviour consistent) or we just default to safe and print a warning message? The latter would mean that the whole Ansible task wouldn't fail at least and the worst thing we've done is leave behind nvram, etc, which we'd notify the user about.

I think either approach works. What do you think? @Andersson007 any thoughts on best approach?

bturmann: Yes, we should catch potential errors. I was already working on it and about to push a new commit.

bturmann:
I just updated the first comment above to reflect the latest overall status.

The PR got a bit larger, but I think for good reasons. With the two new options, it is now possible to `undefine` successfully a guest including nvram or other metadata. Users have a choice whether to simply use `force=true` or provide the desired `flags` individually.

Special cases, like using mutually exclusive flags or if `force` and `flags` are provided together are handled as well with either an exception or warning.

For existing use cases, where none of the new options are provided, then the behaviour is exactly the same as before.

csmart:
@bturmann this looks really great, thanks a lot for the work here! I will do some testing myself and if you're happy, then I'll merge it in.

csmart:
Tested well and looks great to me. Thanks @bturmann!

Andersson007:
@bturmann @csmart thanks folks for your great work!

tomkivlin:
Was this fix released?

I can't see the updated docs on https://docs.ansible.com/ansible/latest/collections/community/libvirt/virt_module.html and when I try with `flags: nvram` I get an error of 

```
Unsupported parameters for (community.libvirt.virt) module: flags. Supported parameters include: autostart, command, name, state, uri, xml (guest).
```

I'm using:
```
community.libvirt             1.2.0      
```

Happy to create an issue if needed but thought I'd check here first.

bturmann:
> Was this fix released?

Indeed, the latest release tag seems to be 1.2.0. I left a comment in the [Release Plan](https://github.com/ansible-collections/community.libvirt/issues/81#issuecomment-1706262757) issue.

csmart:
Thanks for the bump, I will arrange a new release soon.

tomkivlin:
@csmart thanks!  Happy to help (if help is needed).

bturmann:
@tomkivlin
This change is now included in the new release 1.3.0 available on Galaxy. The docs are still showing 1.2.0 and I am not sure when they get updated.

tomkivlin:
Many thanks 🙏

PR #137, opened 2022-08-21
CI: remove f35 from devel

csmart:
Fedora 35 is deprecated from devel, this commit removes it from devel
and adds it to Docker 2_13 so that it continues to be tested.


csmart:
Was waiting to merge #136 to merge first, now rebased and re-testing.

Andersson007:
@csmart thanks for looking after CI!

Issue #144, opened 2022-10-03
Split `virt` into specialized modules


mlow:
##### SUMMARY
Currently `virt` has rather a lot going on.. it's used both to manage resources (libvirt domains) as well as retrieve information of both individual domains of the libvirt node itself (free memory, etc). I think it would be better to split its functionality into a handful of focused modules. These modules could be added without altering any existing functionality. I would image `virt` eventually being deprecated.

##### ISSUE TYPE
- Refactor

##### COMPONENT NAME
- `virt`

##### ADDITIONAL INFORMATION

This is motivated by recent work I've been doing with `community.libvirt` (working towards a clean, declarative way to deploy libvirt VMs, that isn't Terraform, and finding that `community.libvirt` needs some work done to get me there :).

One of the needs I've encountered is the ability to retrieve a domain's interface addresses (also seen in #124). I've got a working prototype of this, but it's challenging to figure out the best way to supply the flag controlling the source of the addressing info (arp, lease, or agent). Meanwhile I can envision a way of implementing this in a separate module which focuses on information retrieval, with a signature of something like:

```yml
---
- community.libvirt.domain_info:
    name: my-domain
    interface_addresses:
      source: agent
```

Which would return something like:
```json
{
  "changed": false,
  "interface_addresses": {
     ... raw libvirt data
  }
}
```

This scheme allows for getting multiple bits of information about a domain at once, e.g.:

```yml
---
- community.libvirt.domain_info:
    name: my-domain
    state:
    interface_addresses:
      source: agent
```

... in a GraphQL-esque fashion, returning data in the same "shape" it's been requested in:

```json
{
  "changed": false,
  "state": "running",
  "interface_addresses": {
     ... raw libvirt data
  }
}
```

The other modules I would expect are:
- `domain` - focusing on domain management (idempotent XML definition and state (running, stopped) management)
- `node_info` - for retrieval of node-wide information (free mem, etc)

I would love for any feedback on the above proposals. If there's agreement that they seem worthwhile, I'll be happy to get started on them.

Cheers

PR #146, opened 2022-10-20
[WIP] Add domain_info

mlow:
##### SUMMARY
Adds a `domain_info` module, specializing in retrieving information about a specific domain.

See #144 for the rational behind this addition.

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
domain_info

##### ADDITIONAL INFORMATION
I'm mainly putting this out there looking for feedback at the moment!

### DIFF
--- a/plugins/modules/domain_info.py
+++ b/plugins/modules/domain_info.py
@@ -1,0 +1,297 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+
+
+DOCUMENTATION = """
+---
+module: domain_info
+short_description: Retrieves information about I(libvirt) domains.
+description:
+     - Retrieves information about I(libvirt) domains.
+options:
+  name:
+    description:
+      - name of the guest domain to get information about.
+    required: true
+  state:
+    description:
+      - If provided, returns the state of the domain I(name)
+    type: dict
+    options: {}
+  xml:
+    description:
+      - If provided, returns the domain's XML.
+    type: dict
+    options:
+      flags:
+        description:
+          - The set of flags to pass to libvirt's GetXMLDesc
+          - See: https://libvirt.org/html/libvirt-libvirt-domain.html#virDomainGetXMLDesc
+        type: list
+        elements: str
+        choices: [ secure, inactive, update-cpu, migratable ]
+  guestinfo:
+    description:
+      - If provided, returns information about the running guest.
+      - Requires qemu-guest-agent on the guest.
+    type: dict
+    options:
+      types:
+        description:
+          - The types of information to get about the guest.
+          - See: https://libvirt.org/html/libvirt-libvirt-domain.html#virDomainGetGuestInfo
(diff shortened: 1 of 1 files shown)


dseeley:
Is it worth considering following the existing names as used for this in the inventory module https://github.com/ansible-collections/community.libvirt/blob/main/plugins/inventory/libvirt.py (`xml_desc`, `guest_info`, `interface_addresses`)?

I also wonder whether `domain_info` is the best name for this module, as `domain.info()` is already an API call, whose results aren't actually included in the information returned by this module, and might be confusing?  Perhaps include the results of `domain.info()` as `info`, (as in `inventory/libvirt.py`), and rename the module to `domain_extended_info `(or something)?

mlow:
> Is it worth considering following the existing names as used for this in the inventory module https://github.com/ansible-collections/community.libvirt/blob/main/plugins/inventory/libvirt.py (xml_desc, guest_info, interface_addresses)?
>
> I also wonder whether domain_info is the best name for this module, as domain.info() is already an API call, whose results aren't actually included in the information returned by this module, and might be confusing? Perhaps include the results of domain.info() as info, (as in inventory/libvirt.py), and rename the module to domain_extended_info (or something)?

I agree it might cause some confusion.. I stumbled on this during writing; the reason I had not (yet) implemented an `info` field is that [domain.info()](https://libvirt.org/html/libvirt-libvirt-domain.html#virDomainInfo) is only a subset of what `dominfo` provides, and my general goal has been to resemble the CLI (which I didn't already with `xml`). That said, I'm happy to rename the fields to match the names in the inventory plugin.

I think we might be able to stick with `domain_info` if we return the values of `domain.info()` by default (including state, which will no longer be an opt-in field). I wouldn't be against renaming it to `dominfo` to match the CLI.

What do you think?

Andersson007 (commented):
@mlow hello, thanks for working on the new module! Nice to see people adding new content to the collection!

Could you please:
- Add `[WIP]` to the PR's description
- Add `version_added:` tag (use the next minor release number)
- Make the docs satisfying the [style guide](https://docs.ansible.com/ansible/latest/dev_guide/style_guide/index.html) (including the paragraph about using macros)
- Be sure the module supports check mode (also please add `notes:` section with `- Supports C(check_mode).`
- Add integration tests. See the content of `tests/integration/targets` and the [integration test quick start guide](https://docs.ansible.com/ansible/devel/community/collection_contributors/collection_integration_tests.html).
- When it's ready for review, remove `[WIP]` from the description

csmart:
Hi @mlow, thanks again for this contribution! I just want to follow up on the items mentioned above, do you think you will have time to update this? Thanks!

mlow:
Hi, sorry to have left this hanging. I do expect I'll have some time to work on this coming up.

My current concern with integration testing is that I'm assuming we can't create KVM domains in the environment (am I wrong?); will it be possible to create LXC domains?

mlow:
Closing this for now. I have a body of work (related to this) that feels like it would fit well into community.libvirt, but currently don't have the time to drive it towards being ready for submission.

PR #162, opened 2023-09-22
fix resolve tilde in file paths (fixes #161)

### LINKED ISSUE #161: Problem with file paths when using ansible.builtin.copy module
##### SUMMARY

Almost all Ansible operations appear to use the unresolved path to the Ansible temp directory '\~/.ansible/tmp'. However, the ansible.builtin.copy module appears to resolve this path to '/root/.ansible/tmp' when looking for the source file. The underlying problem is that when I examined the filesystem on the VM, it appears that the community.libvirt.libvirt_qemu connection plugin is copying all files in the temp directory to '/\~/.ansible/tmp'. The tilde is not being resolved to the homedir of the user and instead ends up in a '/\~' directory at the root of the filesystem. 

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

community.libvirt.libvirt_qemu connection plugin

##### ANSIBLE VERSION

```paste below
ansible [core 2.15.4]
  config file = /home/user/ansible-dir/ansible.cfg
  configured module search path = ['/home/user/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/user/ansible-dir/.venv/lib/python3.11/site-packages/ansible
  ansible collection location = /home/user/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/user/ansible-dir/.venv/bin/ansible
  python version = 3.11.5 (main, Aug 25 2023, 13:19:50) [GCC 11.4.0] (/home/user/ansible-dir/.venv/bin/python)
  jinja version = 3.1.2
  libyaml = True
```

##### COLLECTION VERSION

```paste below
# /home/user/ansible-dir/.venv/lib/python3.11/site-packages/ansible_collections
Collection        Version
----------------- -------
community.libvirt 1.2.0  
```

##### CONFIGURATION

```paste below
CONFIG_FILE() = /home/user/ansible-dir/ansible.cfg
DEFAULT_HOST_LIST(/home/user/ansible-dir/ansible.cfg) = ['/home/user/ansible-dir/inventory.yml']
DEFAULT_KEEP_REMOTE_FILES(env: ANSIBLE_KEEP_REMOTE_FILES) = False
EDITOR(env: EDITOR) = /home/user/bin/vim
```

##### OS / ENVIRONMENT

Target OS: Ubuntu 22.04 Server

##### STEPS TO REPRODUCE

```yaml
ansible -i local-libvirt-inventory.yml --become all -m copy -a 'src=srcfile dest=/path/to/dest/file'
```

##### EXPECTED RESULTS

The local file is copied to the libvirt host.

##### ACTUAL RESULTS

The file is not copied. When examining the filesystem in the libvirt VM, I found that the directory '/root/.ansible/tmp' doesn't exist. It is located at '/\~/.ansible/tmp' instead.

```paste below
libvirt-host | FAILED! => {
    "changed": false,
    "checksum": "fac0b7e52fc298c5c5feb6b7673f74e2306aba6e",
    "msg": "Source /root/.ansible/tmp/ansible-tmp-1695394376.1436307-368511-193439908275987/source not found"
}
```

antonc42:
##### SUMMARY

The libvirt_qemu connection module uses QEMU guest agent to copy files
and run commands on the remote host. Ansible uses the '\~/' symbolic path
to the user's home directory for many commands and file paths.
Unfortunately, since QEMU guest agent doesn't run in a shell, it doesn't
resolve the '\~/' path to the user's homedir.

Instead, the Ansible paths end up at '/\~/' in the remote host. This
breaks some assumptions for some modules about the path to the Ansible
files. Some modules use the resolved homedir path and fail to find any
files in that location.

This fix first resolves the remote user homedir with a QEMU guest
command, caches the result, and then replaces '\~/' found in any command
or remote file paths with the resolved absolute path to the user's
homedir.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

community.libvirt.libvirt_qemu connection plugin

##### ADDITIONAL INFORMATION

I've tested this patch with playbooks that use many different Ansible modules and it seems to work in all the ones I tested, including the copy module that was having issues in #161.

### DIFF
--- a/changelogs/fragments/161_fix_resolve_tilde.yml
+++ b/changelogs/fragments/161_fix_resolve_tilde.yml
@@ -1,0 +1,2 @@
+bugfixes:
+  - libvirt_qemu - fix path resolution of tilde (~) used to represent remote user's homedir
--- a/plugins/connection/libvirt_qemu.py
+++ b/plugins/connection/libvirt_qemu.py
@@ -46,6 +46,7 @@ options:
 
 import base64
 import json
+import re
 import shlex
 import time
 import traceback
@@ -98,6 +99,7 @@ class Connection(ConnectionBase):
         super(Connection, self).__init__(play_context, new_stdin, *args, **kwargs)
 
         self._host = self._play_context.remote_addr
+        self._user_homedir = None
 
         # Windows operates differently from a POSIX connection/shell plugin,
         # we need to set various properties to ensure SSH on Windows continues
@@ -146,6 +148,21 @@ class Connection(ConnectionBase):
             display.vvv(u"ESTABLISH {0} CONNECTION".format(self.transport), host=self._host)
             self._connected = True
 
+    @property
+    def user_homedir(self):
+        """ the resolved user homedir on the remote """
+        if self._user_homedir:
+            return self._user_homedir
+        exitcode, stdout, stderr = self.exec_command("/bin/sh -c 'getent passwd $(id -un) | cut -d: -f6'")
+        self._user_homedir = to_text(stdout).strip()
+        return self._user_homedir
+
+    def _resolve_tilde(self,string):
+        """ resolve file paths or commands that begin with '~/' to the remote user's homedir """
+        if re.search(r"~\/",string):
+            return re.sub(r"~\/", self.user_homedir + r"/",string)
+        return string
+
     def exec_command(self, cmd, in_data=None, sudoable=True):
         """ execute a command on the virtual machine host """
         super(Connection, self).exec_command(cmd, in_data=in_data, sudoable=sudoable)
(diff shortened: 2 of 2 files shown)


leegarrett:
This looks good to me, it fixes the issue #161. Can we get this merged?

csmart:
Hi @antonc42 thanks for the contribution! I've been trying to replicate this and finally did on Ubuntu, see my comment in the issue here https://github.com/ansible-collections/community.libvirt/issues/161#issuecomment-1913109726

I think it may be a problem with `dash` being the default shell, it seems to work as expected if I switch the default shell to `bash`. Would appreciate if you can see if that also works for you.

In terms of this patch, I would prefer to see if there's another way to solve it rather than using specific Linux commands, however if we do, then we may need some if checks for the OS as I'm not sure that the patch as it is would work on Windows guests.

/cc @leegarrett

antonc42:
@csmart
I'm not sure if it's possible to use `os.path.expanduser` since that would be executed on the Ansible controller instead of in the guest VM, correct? The user homedir inside the VM is needed to resolve the tilde. The only way I knew of to interact with the guest VM is through commands. Is there another way?

I suppose I could execute a Python command on the guest VM, e.g. `python3 -c 'import os; print(os.path.expanduser("~"))'`. I'm not sure if that is universal enough to run on Windows guests as I don't have any experience with Python on Windows.

antonc42:
Not needed. See [my comment](https://github.com/ansible-collections/community.libvirt/issues/161#issuecomment-2289213882) in the original issue #161.

PR #169, opened 2023-11-22
ci: swap rhel 9.2 with 9.3

csmart:
RHEL 9.2 has been deprecated from ansible-test and replaced with RHEL 9.3. This change applies this to the devel tests.


csmart:
closing in favour of !170 as there were more things to fix.

PR #180, opened 2024-10-24
Implement ansible module for management of libvirt volumes

NK308:
##### SUMMARY

This PR adds a module to create, delete and modify libvirt volumes.

This PR is based on the fork which was subject of https://github.com/ansible-collections/community.libvirt/pull/45

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

`virt_volume`

##### ADDITIONAL INFORMATION

Some of the code of the new ansible module as well as parts of the code of `virt_pool`, have been refactored to `module_utils` to reduce redundancy.
For now I have only tested the simple creation und deletion of storage volumes, so there might some more testing be needed.
For now I didn't do much more, compared to https://github.com/ansible-collections/community.libvirt/pull/45, than bringing the branch up to date with the main branch, including the adoption of changes in the code and doc style from the other modules.


NK308:
Okay ... I don't understand, why this error is so specific for RHEL.

NK308:
The integration tests now fail because of  ... keyboard interrupt? How?

Andersson007:
@NK308 hello, i wonder if we can remote those remote targets

plugins/modules/virt_volume.py
```diff
@@ -0,0 +1,507 @@
...
+ALL_MODES.extend(ENTRY_DELETE_FLAGS_MAP.keys())
+
+from ansible_collections.community.libvirt.plugins.module_utils.pool import LibvirtConnection as PoolConnection, HAS_VIRT, HAS_XML
+from ansible_collections.community.libvirt.plugins.module_utils.entry import EntryNotFound
+
+if HAS_VIRT:
+    import libvirt
+
+
+class LibvirtConnection(object):
```
Andersson007, line 192:
```suggestion
class LibvirtConnection(object):
```
don't you import this class from module_utils.pool?

plugins/modules/virt_volume.py
```diff
@@ -0,0 +1,507 @@
...
+    "normal": 0,
+    "zeroed": 1
+}
+
+ALL_MODES = []
+ALL_MODES.extend(ENTRY_BUILD_FLAGS_MAP.keys())
+ALL_MODES.extend(ENTRY_DELETE_FLAGS_MAP.keys())
+
+from ansible_collections.community.libvirt.plugins.module_utils.pool import LibvirtConnection as PoolConnection, HAS_VIRT, HAS_XML
+from ansible_collections.community.libvirt.plugins.module_utils.entry import EntryNotFound
```
Andersson007, line 186:
```suggestion
from ansible_collections.community.libvirt.plugins.module_utils.pool import LibvirtConnection as PoolConnection, HAS_VIRT, HAS_XML
from ansible_collections.community.libvirt.plugins.module_utils.entry import EntryNotFound
```
I think this should be moved up to line 131

plugins/modules/virt_volume.py
```diff
@@ -0,0 +1,507 @@
...
+
+if HAS_VIRT:
+    import libvirt
```
Andersson007, line 189:
```suggestion
if HAS_VIRT:
    import libvirt
```
isn't it already imported?

Andersson007 (commented):
@NK308 thanks for the PR
I would separate the code:
1. You refactor the code in one PR
2. Then you add the module in another PR or wise versa

It's easy to make a mistake and imo not a good idea in general to mix refactoring and new features in a single PR.
Hence, i'm suggesting splitting the PR.

cc @csmart

NK308:
> @NK308 thanks for the PR I would separate the code:
> 
>     1. You refactor the code in one PR
> 
>     2. Then you add the module in another PR or wise versa
> 
> 
> It's easy to make a mistake and imo not a good idea in general to mix refactoring and new features in a single PR. Hence, i'm suggesting splitting the PR.
> 
> cc @csmart

Hi
I the reason, why the new module and the refactoring are mixed up here is, that I based my branch on a pre-existing branch, which came with this intermixing of new features and refactoring, and I tried to keep changes to a minimum while resolving the conflicts between that branch and main.
I probably would have done some things different, if I wrote the module all by myself.

plugins/modules/virt_volume.py
NK308: The  imported class is not called `LibvirtConnection` in the namespace of `plugin.modules.libvirt` which is something I adopted from https://github.com/flagbot/community.libvirt/tree/feature/virt_volume.

plugins/modules/virt_volume.py
NK308: Not in THIS module. The bool `HAS_VIRT` has been importet from another module, which has tried to import `libvirt` and with this, I'm basically avoiding, to have to raise and catch the `ImportError` again.

NK308:
About the refactoring topic, if I'm gonna split the refactoring off this PR anyway...
This collection seems to contain a lot of redundant code since each module is basically a wrapper around a class from the libvirt python bindings. Maybe all of the modules should be refactored together, introducing some inheritance?

The coding style seems to be a bit weird in general. The `find_entry()` function for example is two completely different functions in one:
``
@overload
def find_entry(self, entryid: str) -> virSomething
def find_entry(self, entryid: int = -1) -> list[virSomething]
``So there is actually no reason, to not split that into two different functions.
Also the naming is a bit irritating since there in most modules, there are at least two classes available with a function named `find_entry`, which are not related but basically do the same, with different return types. Having the same name would make sense if the classes inherited the method from the same super class, but since they don't, names like `find_volume`, `find_pool`, `find_domain`, etc. would be a lot more self-explanatory.

Also it seems very arbitrary, which methods take the python wrapper of some libvirt object, or it's name (and call `find_entry` by itself to get the wrapper). I think, it would be better coding style, to have a wrapper class for all the classes from the libvirt python-bindings, so that each of the libvirt objects gets wrapped into its own object instead of writing classes, consisting only of methods, which could as well be standalone functions.

I'm not sure, how much of that might be prevented by the requirement to keep compatibility down to python2. Things like generic classes surely won't work.

NK308:
> @NK308 hello, i wonder if we can remote those remote targets

I'm not sure about that. In some of the previous commits, there was an actual error message, which showed an actual problem with an integration test, which dissappeared after I fixed it. And this error only showed on the Remote RHEL targets for  some reason, I don't understand.

NK308:
"no connection driver available for qemu:///system" is a now error, and I'm not able to tell, why it starts showing up exactly now, and not earlier.

Andersson007:
I'm unfortunately completely unfamiliar with the technology and have no idea about why the tests fail
@csmart @drybjed @odyssey4me any thoughts?

Andersson007:
@NK308 i took a quick look, i see you test on very old distros, how about just testing only on ubuntu 24.04 if it makes sense?

Andersson007:
> @Andersson007 The issue seems to be very specific to the RHEL remote targets, even the current version, while older versions of other operating systems don't seem to cause any problems. I was also assuming, a PR has to pass those targets to be accepted to a collection in the community namespace.

@NK308 my comment is not related to the failures, i just think it's an extra thing and there's no such a requirement to test against them. I'm not familiar with the technology though and I don't know if it's distro dependent or not.
If it is, it makes sense to test against several relevant distros but I guess it doesn't make sense to test on not supported ones.

NK308:
I now have removed the OS specific variables from the integration tests, if that's what you meant. The remaining operating systems represented in `tests/integration/targets/virt_volume/vars` are actually not yet end of life, despite being old, or at least still receive some kind of extended support.

Andersson007:
@NK308 how about removing RH 7 and updating Ubuntu 16 to 24.04 ?

csmart:
Hi @NK308 thanks very much for reviving this, I'll take a look and put in any comments in line.

csmart (commented):
@NK308 thanks again for this, I've done an initial review with comments in line, if you have time to take a look that would be great. I think one of the things we should consider is perhaps combining the volume and pool tests as volume is dependent on a pool, or reworking them to have some common parts to call. Also, I think it'd be great if the format of the volume tests were similar in style to the pool tests, e.g. with "check" mode, etc. I did test check mode but it didn't seem to behave as expected, so that might need some attention (but that's an inline comment). Feel free to let me know how you go and if you need a hand. Thanks!

NK308:
The original branch from @flagbot on which my branch is based, had shared code between the `virt_storage` and `virt_pool` modules, but @Andersson007 recommended me to put anything, that involves the total refactoring of an already existing module, into a separate PR. That's why I basically reversed the refactoring that exists in https://github.com/flagbot/community.libvirt/tree/feature/virt_volume.

Andersson007:
> The original branch from @flagbot on which my branch is based, had shared code between the `virt_storage` and `virt_pool` modules, but @Andersson007 recommended me to put anything, that involves the total refactoring of an already existing module, into a separate PR. That's why I basically reversed the refactoring that exists in https://github.com/flagbot/community.libvirt/tree/feature/virt_volume.

@NK308 can we incorporate any of the suggestions by @csmart or you think it's sufficient for merge as is? I can't provide any feedback on the refactoring but it'm for keeping unrelated things separate (i.e. in separate PRs).

NK308:
> > The original branch from @flagbot on which my branch is based, had shared code between the `virt_storage` and `virt_pool` modules, but @Andersson007 recommended me to put anything, that involves the total refactoring of an already existing module, into a separate PR. That's why I basically reversed the refactoring that exists in https://github.com/flagbot/community.libvirt/tree/feature/virt_volume.
> 
> @NK308 can we incorporate any of the suggestions by @csmart or you think it's sufficient for merge as is? I can't provide any feedback on the refactoring but it'm for keeping unrelated things separate (i.e. in separate PRs).

I think, the best thing would be, to incorporate some of the smaller changes, merge this PR, then do some refactoring in a second PR and then handling the stuff about the diffs and checks in a third PR.

But thinking about the diffs... I can think about some features, that would have some overlapp with that topic, like a command which would be to able to edit/update things like domains or networks. I mean ... both features would require code to do a more in-depth analysis of the xml definitions, that I can find here right now. This might hold for the freatures of https://github.com/ansible-collections/community.libvirt/pull/183 (I didn't really take a look into it jet).

So before we do the refactoring, we might want to have a more in-depth discussion about how we want the outline of the code in module_utils to look like, and how it should look like, to provide a good foundation for new features in the future.
At least to me, the point, where we start to refactor common code into module_utils, seems to be the right point in time for this discussion.

plugins/modules/virt_volume.py
```diff
@@ -0,0 +1,711 @@
...
+
+    if not HAS_XML:
+        module.fail_json(
+            msg='The `lxml` module is not importable. Check the requirements.'
+        )
+
+    rc = VIRT_SUCCESS
+    try:
+        rc, result = core(module)
+    except Exception as e:
```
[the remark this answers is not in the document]
NK308: For now, it's probably enough to use `EntryNotFound`and since Volumes are also entries. and then define more specific Exceptions during the refactoring maybe?

plugins/modules/virt_volume.py
```diff
@@ -0,0 +1,711 @@
...
+---
+module: virt_volume
+author:
+    - Leonardo Galli (@galli-leo)
+short_description: Manage libvirt volumes inside a storage pool
+description:
+    - Manage I(libvirt) volumes inside a storage pool.
+options:
+    name:
+        aliases: [ "volume" ]
```
Andersson007, line 21:
```suggestion
```
Only `name` is ok here, there's a recommendation in the docs to avoid aliases whenever possible

plugins/modules/virt_volume.py
```diff
@@ -0,0 +1,711 @@
...
+            - Name of the storage pool, where the volume is located.
+        type: str
+    state:
+        choices: [ "present", "absent", "undefined", "deleted" ]
+        description:
+            - Specify which state you want a volume to be in.
+              If 'present', ensure that the volume is present but do not change its
+              state; if it's missing, you need to specify xml argument.
+              If 'undefined' or 'absent', volume will be removed from I(libvirt) configuration (logically only!).
+              If 'deleted', volume will be wiped clean and then removed.
```
Andersson007, line 38:
```suggestion
            - If C(present), ensure that the volume is present but do not change its
              state; if it's missing, you need to specify xml argument.
            - If C(undefined) or C(absent), volume will be removed from I(libvirt) configuration (logically only!).
            - If C(deleted), volume will be wiped clean and then removed.
```
should we also explain a difference between undefined and absent?
If they do the same, should we leave only one of them?

plugins/modules/virt_volume.py
```diff
@@ -0,0 +1,711 @@
...
+            module.fail_json(msg="Command %s not recognized" % command)
+
+    module.fail_json(msg="expected state or command parameter to be specified")
+
+
+def main():
+
+    module = AnsibleModule(
+        argument_spec=dict(
+            name=dict(aliases=['volume']),
```
Andersson007, line 677:
```suggestion
            name=dict(),
```

plugins/modules/virt_volume.py
```diff
@@ -0,0 +1,711 @@
...
+    - Leonardo Galli (@galli-leo)
+short_description: Manage libvirt volumes inside a storage pool
+description:
+    - Manage I(libvirt) volumes inside a storage pool.
+options:
+    name:
+        aliases: [ "volume" ]
+        description:
+            - Name of the volume being managed. Note that the volume must be previously
+              defined with xml.
```
Andersson007, line 24:
```suggestion
            - Name of the volume being managed. Note that the volume must be previously
              defined with xml.
```
Is it not required? Just to be sure

plugins/modules/virt_volume.py
```diff
@@ -0,0 +1,711 @@
...
+
+
+DOCUMENTATION = '''
+---
+module: virt_volume
+author:
+    - Leonardo Galli (@galli-leo)
+short_description: Manage libvirt volumes inside a storage pool
+description:
+    - Manage I(libvirt) volumes inside a storage pool.
```
Andersson007, line 18:
```suggestion
    - Manage I(libvirt) volumes inside a storage pool.
attributes:
    check_mode:
        description: Supports check_mode.
        support: full
```

plugins/modules/virt_volume.py
```diff
@@ -0,0 +1,711 @@
...
+# (c) 2015, Maciej Delmanowski <drybjed@gmail.com>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+
+DOCUMENTATION = '''
+---
+module: virt_volume
```
Andersson007, line 13:
```suggestion
module: virt_volume
version_added: '1.4.0'
```

Andersson007 (commented):
Cool, let's proceed, I can help with a general review, there we go

Andersson007:
I would highly suggest updating ubuntu 16 to at least 22 or better 24 and RedHat to the latest. Those old ones are really old and receive only security updates

plugins/modules/virt_volume.py
NK308: I think, the aliases are a good idea here, because in playbooks with lots of tasks from `community.libvirt` it improves the readability, to not have every task contain a `name` which references to a totally different kind of structure.

plugins/modules/virt_volume.py
NK308: Yes, it seems to be totally the same. I'm gonna remove `undefined` since the other one sounds more fitting for the vocabulary of ansible.

plugins/modules/virt_volume.py
NK308: It is required, with the exception of the `list_volumes`, `info` and `facts` command.

csmart:
> The original branch from @flagbot on which my branch is based, had shared code between the `virt_storage` and `virt_pool` modules, but @Andersson007 recommended me to put anything, that involves the total refactoring of an already existing module, into a separate PR. That's why I basically reversed the refactoring that exists in https://github.com/flagbot/community.libvirt/tree/feature/virt_volume.

Different MR for unrelated features is good, but I'm also wary of merging something for which the coverage of the tests isn't quite complete and then it never gets fixed up later. Duplication is fine, but I think the `check` is something that probably should be fixed here to help create a complete test. But if @Andersson007 prefers to merge it as it is, then we can work on that and try to clean it up later.

csmart:
Just wanted to follow this up to see if we were thinking we should merge this as is, and then follow up with a separate MR? Thanks!

Andersson007:
> Just wanted to follow this up to see if we were thinking we should merge this as is, and then follow up with a separate MR? Thanks!

I'm OK with either option

csmart:
@stratus-ss are you interested in having a look at this MR too? I'm interested in what you think and also that it complements #191

stratus-ss:
> @stratus-ss are you interested in having a look at this MR too? I'm interested in what you think and also that it complements #191

I'll take a deeper look shortly. I read this over a few times and nothing specific jumped out at me. I'd have to think through how this might be achieved in order to reflect on what the user has done here.

Having said all that, I doubt I will come up with anything near what @NK308 has done here

csmart:
Just following up whether we think we should merge this as is, as per above comments.. thanks!

csmart:
Let's merge this as it is and hope we can fix any shortcomings in the future, given it's been dragging on a while. Thanks @NK308

PR #183, opened 2024-12-01
Implement --check support for the libvirt.virt module

### LINKED ISSUE #98: ansible-playbook --check mode/diff support for virt module
##### SUMMARY

The `virt` module does not seem to support ansible-playbook `--check` mode - in [check](https://docs.ansible.com/ansible/latest/user_guide/playbooks_checkmode.html) mode, tasks that use the virt module will always return `skipped`.

##### ISSUE TYPE

- Feature Idea

##### COMPONENT NAME

virt

##### ADDITIONAL INFORMATION

This would allow reviewing changes in "dry-run" mode before actually applying them. Running the `virt` module in check mode should return the expected `ok/changed` state, and a diff of the VM XML definition when relevant (for example when `command: define`).

```yaml
ansible-playbook playbook.yml --tags libvirt --check --diff
```

lowjoel:
##### SUMMARY

Implement --check support for the libvirt.virt module

Fixes #98.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

virt

##### ADDITIONAL INFORMATION

Initial implementation to support --check mode.

This adds checking to the Virt class to make check mode calls no-ops but returning valid information where possible. For the get methods this doesn't affect behaviour, but for state changes and for defining domains this will make the Virt class do nothing in check mode.

For defining a new domain, this will pretty-print both XML definitions to make the diff output clean. It's not perfect because libvirt adds additional attributes when defining a domain that are implicit. Those will not change the domain if they are not specified in the input XML.

Tested check mode: define/undefine.

```
TASK [libvirt-domain : Define Domain (domain)] ***********************************************************************************************************************************************************************************************************************************************
--- before
+++ after
@@ -1,37 +1,37 @@
-<domain type="kvm">
+<domain xmlns:qemu="http://libvirt.org/schemas/domain/qemu/1.0" type="kvm">
   <name>domain</name>
   <uuid>snip</uuid>
-  <title>Windows 11</title>
-  <description>Windows 11 21H2</description>
+  <title>Windows 10</title>
+  <description>Windows 10 21H2</description>
   <metadata>
     <libosinfo:libosinfo xmlns:libosinfo="http://libosinfo.org/xmlns/libvirt/domain/1.0">
...snip
```

### DIFF
--- a/tests/integration/targets/virt/tasks/main.yml
+++ b/tests/integration/targets/virt/tasks/main.yml
@@ -1,0 +1,31 @@
+---
+- include_vars: '{{ item }}'
+  with_first_found:
+    - "{{ ansible_distribution }}-{{ ansible_distribution_version}}.yml"
+    - "{{ ansible_distribution }}-{{ ansible_distribution_major_version}}.yml"
+    - "{{ ansible_distribution }}.yml"
+    - "default.yml"
+
+- block:
+    - name: Install libvirt packages
+      package:
+        name: "{{ virt_domain_packages }}"
+
+    - name: Start libvirt service
+      service:
+        name: libvirtd
+        state: started
+
+    - name: Test check mode for domain defines
+      import_tasks: define_check_mode.yml
+
+  always:
+    - name: Stop libvirt
+      service:
+        name: libvirtd
+        state: stopped
+
+    - name: Remove only the libvirt packages
+      package:
+        name: "{{ virt_domain_packages|select('match', '.*libvirt.*')|list }}"
+        state: absent
--- a/tests/integration/targets/virt/templates/test_domain.xml.j2
+++ b/tests/integration/targets/virt/templates/test_domain.xml.j2
@@ -1,0 +1,51 @@
+<domain type='qemu'>
+  <name>{{ domain_info['name'] }}</name>
+  <uuid>125388f2-6ab4-4343-a166-bad6ffad017e</uuid>
+  <memory unit='KiB'>4000768</memory>
+  <currentMemory unit='KiB'>4000000</currentMemory>
+  <vcpu placement='static'>1</vcpu>
+  <os>
+    <type arch='x86_64' machine='pc-i440fx-{{ emulator_version }}'>hvm</type>
+    <boot dev='hd'/>
+  </os>
+  <features>
+    <acpi/>
+    <apic/>
+    <pae/>
+  </features>
+  <cpu mode='custom' match='exact' check='none'>
+    <model fallback='forbid'>qemu64</model>
+  </cpu>
+  <clock offset='utc'>
+    <timer name='kvmclock' present='no'/>
+  </clock>
+  <on_poweroff>destroy</on_poweroff>
+  <on_reboot>restart</on_reboot>
+  <on_crash>restart</on_crash>
+  <devices>
+    <emulator>{{ emulator_bin }}</emulator>
+    <controller type='usb' index='0' model='piix3-uhci'>
+      <address type='pci' domain='0x0000' bus='0x00' slot='0x01' function='0x2'/>
+    </controller>
+    <controller type='pci' index='0' model='pci-root'/>
+    <controller type='pci' index='1' model='pci-bridge'>
+      <model name='pci-bridge'/>
+      <target chassisNr='1'/>
+      <address type='pci' domain='0x0000' bus='0x00' slot='0x02' function='0x0'/>
+    </controller>
+    <controller type='pci' index='2' model='pci-bridge'>
+      <model name='pci-bridge'/>
+      <target chassisNr='2'/>
+      <address type='pci' domain='0x0000' bus='0x00' slot='0x03' function='0x0'/>
+    </controller>
+    <controller type='pci' index='3' model='pci-bridge'>
+      <model name='pci-bridge'/>
+      <target chassisNr='3'/>
+      <address type='pci' domain='0x0000' bus='0x00' slot='0x04' function='0x0'/>
+    </controller>
+    <input type='mouse' bus='ps2'/>
+    <input type='keyboard' bus='ps2'/>
+    <audio id='1' type='none'/>
+    <memballoon model='none'/>
+  </devices>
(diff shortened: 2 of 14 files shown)


Andersson007:
@lowjoel hello, thanks for your effort! please also add a changelog fragment of minor_changes type https://docs.ansible.com/ansible/latest/community/development_process.html#creating-a-changelog-fragment

lowjoel:
Thanks! I'm not entirely sure this is ready to merge yet; I think the state changes aren't reported as changes (not sure)

csmart:
hi @lowjoel thanks for this contribution! Wondering whether it's possible to add some tests to validate the functionality, like is done with `virt_pool`? I think it'd be great to make sure it's behaving as expected. Cheers!

lowjoel:
Pretty much duplicated it and replaced it with `domain`, is there some CI infrastructure I can test that it all works? I could figure out how to test this out using Docker or the like but it'll need some rummaging through docs and could take a while.

Andersson007:
> Pretty much duplicated it and replaced it with `domain`, is there some CI infrastructure I can test that it all works? I could figure out how to test this out using Docker or the like but it'll need some rummaging through docs and could take a while.

@lowjoel hello, please take a look at
- https://docs.ansible.com/ansible/devel/community/create_pr_quick_start.html#test-your-changes 
- (if a test role is not there yet which you'll figure out from ^) https://docs.ansible.com/ansible/devel/community/collection_contributors/collection_integration_add.html

lowjoel:
> @lowjoel hello, please take a look at
thanks! though in my case the harder part is to have a libvirt instance to test with because I (obviously) shouldn't be testing this in my production setup.

lowjoel:
@Andersson007 @csmart Ran the tests locally. Works well enough -- tested that the define/undefine cycle works fine. I can't fully verify the state change because I can't get qemu to start in a container. But otherwise it should work fine now.

Andersson007:
@lowjoel I think the PR is fine to be merged now. One small thing: could you please add the following lines to the DOCUMENTATION section of the module file so that users know check_mode is supported:
```
attributes:
  check_mode:
    description: Supports check_mode.
    support: full
```

lowjoel:
@Andersson007 done, hope this is what you meant.

Andersson007 (approved):
+1 to merge (but I've no idea about the technology specific stuff including the CI0
CC @csmart

Andersson007:
@lowjoel I've just ping Chris directly. If no feedback, I'll merge the PR within days, otherwise Chris will
December / early January is a busy time for everyone I believe, sorry for the delay

lowjoel:
no worries - let me know if there's anything I should fix 😀 happy to help

tests/integration/targets/virt/tasks/main.yml
```diff
@@ -0,0 +1,31 @@
+---
+- include_vars: '{{ item }}'
+  with_first_found:
+    - "{{ ansible_distribution }}-{{ ansible_distribution_version}}.yml"
+    - "{{ ansible_distribution }}-{{ ansible_distribution_major_version}}.yml"
+    - "{{ ansible_distribution }}.yml"
+    - "default.yml"
```
csmart, line 7: could we also add `{{ ansible_os_family }}` and `{{ ansible_os_family }}-{{ ansible_distribution_major_version }}` checks here before `default` here? It might make it easier to test things like Rocky without having to duplicate the yml files unnecessarily.

tests/integration/targets/virt/vars/Fedora-34.yml
csmart, line 1: could we drop the version here to support all Fedora releases, we can add a specific one if the packages change

tests/integration/targets/virt/vars/RedHat-7.yml
csmart, line 1: I think this could be dropped now, as it's deprecated

tests/integration/targets/virt/vars/RedHat-8.yml
csmart, line 1: if RHEL 8 and 9 are the same, we could drop this 8 yml file and rename `RedHat-9.yml` to just `RedHat.yml`

tests/integration/targets/virt/templates/test_domain.xml.j2
```diff
@@ -0,0 +1,51 @@
+<domain type='qemu'>
+  <name>{{ domain_info['name'] }}</name>
+  <uuid>125388f2-6ab4-4343-a166-bad6ffad017e</uuid>
```
csmart, line 3: I think we could prob drop `uuid` and it will get generated automatically when the VM is defined, this will at least keep it unique with each test run in case they are run in parallel somehow down the track

tests/integration/targets/virt/vars/Ubuntu-24.04.yml
csmart, line 1: this could prob be renamed to just `Ubuntu` as it's likely newer releases will be the same, and will prevent us from needing to add a new file just to get test working

csmart:
@lowjoel this is looking good, thanks for the tests - I will try to find some time shortly to manually test those, in the mean time I've added a couple of small suggestions

Andersson007:
@lowjoel any comments on @csmart 's statements above?

lowjoel:
@Andersson007 @csmart Addressed your comments, please have a look again 🙇

tests/integration/targets/virt/templates/test_domain.xml.j2
lowjoel: This would cause the first test for "not changed" to fail - since the diff is really a syntactic diff (rather than semantic diff)

Andersson007:
@lowjoel hi, please rebase and solve the conflict, there's one small as far as i can see
then we could merge i think

Andersson007:
@lowjoel thanks for the contribution!
@csmart thanks for reviewing!

@lowjoel if you take a look at a couple of PRs in the queue https://github.com/ansible-collections/community.libvirt/pulls and say what you think as a collection user/developer, it'd be much appreciated!

PR #199, opened 2025-06-13
fix examples for virt_volume

csmart:
The examples provided in virt_volume were copied from virt_pool and not relevant to the module.


Andersson007:
@csmart thanks for the contribution!

PR #214, opened 2025-08-27
refactor: Extract common virt-install code to module_utils and doc_fragments

thinkdoggie:
##### SUMMARY

This refactoring enables code reuse between virt_install and virt_cloud_instance modules while leveraging shared doc_fragments for consistent documentation and fixing several parameter type issues that prevented proper dict-based configuration options from working correctly.

- Move VirtInstallTool class and related utilities from virt_install module module_utils
- Refactor virt_install module to use doc_fragments for module documentation
- Fix parameter type handling for metadata, events, resource, and sysinfo options
- Fix cloud_init.meta_data parameter type from dict to str in doc_fragments
- Update virt_cloud_instance module to use refactored utilities with cleaner imports

##### ISSUE TYPE
- Bugfix Pull Request
- Feature Pull Request

##### COMPONENT NAME
`virt_install`
`virt_cloud_instance`


bturmann:
The approach of splitting code in smaller pieces in module utils as shared resources is certainly a good idea. Also, there are tests included, so the risks are reduced. I hope, that we can cover all modules with tests and refactor them in a similar fashion.

That said, I am not so sure, if it is a good idea to introduce functionality based on other external dependencies, like virt-install in this case. I rather prefer a native libvirt implementation based on that API and I am fully aware that this is only possible by providing configuration based on XML files which is a real obstacle. Means, I fully understand the approach taken with virt-install as it makes things much easier.

Ideally, the user provides configuration with standard ansible variables in YAML and the heavy lifting with XML is done internally with the libvirt API hidden to the user.

thinkdoggie:
> The approach of splitting code in smaller pieces in module utils as shared resources is certainly a good idea. Also, there are tests included, so the risks are reduced. I hope, that we can cover all modules with tests and refactor them in a similar fashion.
> 
> That said, I am not so sure, if it is a good idea to introduce functionality based on other external dependencies, like virt-install in this case. I rather prefer a native libvirt implementation based on that API and I am fully aware that this is only possible by providing configuration based on XML files which is a real obstacle. Means, I fully understand the approach taken with virt-install as it makes things much easier.
> 
> Ideally, the user provides configuration with standard ansible variables in YAML and the heavy lifting with XML is done internally with the libvirt API hidden to the user.

Hi @bturmann , thank you for your thoughtful feedback on the architectural approach. I appreciate your perspective on maintaining consistency with other modules and reducing external dependencies. Let me address your concerns about using virt-install versus a pure libvirt API implementation.
#### virt-install as Part of the Libvirt Ecosystem
While I understand the preference for native libvirt API usage, virt-install should be considered an integral part of the libvirt ecosystem rather than an external dependency. It's distributed alongside libvirt and virsh in all major distributions and serves as the standard tool for VM creation and installation.
#### Semantic Mapping Considerations
The key challenge lies in the semantic gap between Ansible's declarative YAML and libvirt's XML configuration. Consider this example:
Ansible YAML: ```memory: 2048, vcpus: 2, os: fedora39```
virt-install: ```--memory 2048 --vcpus 2 --os-variant fedora39``` (1:1 semantic mapping)
libvirt XML: Requires complex transformation involving memory unit conversion, CPU topology decisions, machine type selection, driver choices, and OS-specific optimizations
virt-install parameters are designed to bridge user intent with hypervisor capabilities, while libvirt XML represents low-level hypervisor configuration. This makes virt-install a natural fit for Ansible's declarative approach.
#### Engineering Complexity
From a practical standpoint, reimplementing virt-install's functionality would involve porting tens of thousands of lines of complex Python code. This would also introduce additional Python package dependencies (libosinfo, requests, etc.) that would complicate deployment and maintenance. Our current implementation maintains "lean and mean" principle with a lightweight wrapper that focuses on parameter translation rather than reimplementing business logic.
#### Conclusion
While I agree that code reusability and reduced dependencies are important architectural principles, the complexity and specialized nature of VM provisioning makes leveraging virt-install the pragmatic choice. It provides users with a familiar YAML interface while ensuring reliable, optimized VM creation through battle-tested logic.
I'm open to discussing this further or exploring hybrid approaches if you have specific concerns about the current implementation.

thinkdoggie:
> > I've come up with a new approach to address this incompatibility issue. We can define the three suboptions (meta_data, user_data, network_config) as `raw` type and allow either a str or dict input value.
> > Do you think if it can prevent the breaking changes for the 2.0.0 release?
> 
> @thinkdoggie it sounds like a solution to me, thanks. Ping us explicitly again once the new implementation is ready for review, thanks. Also if you change you mind about this approach (e.g. it's too hard to implement, etc), please let us know too (I'm writing "us" hoping there'll be more participants/reviewers in the discussion)

Hi @Andersson007, I implemented the new solution and updated the changelog fragment. Please kindely review the new commit. If you prefer, I can squash two commit into one in this PR. Thanks!

thinkdoggie:
Hi @Andersson007, I updated the changelog fragment and squashed two commits into one. I also did E2E regression tests for the virt_install module and virt_cloud_instance_model. Thanks for your reviewing!

In addition, could you help check why the "CI (Sanity & Units devel Sanity)" and "CI" checks failed?

Andersson007:
@thinkdoggie , merged, thanks for the contribution!
The sanity checks failures don't look related. They should be fixed. As they happen when we test against `devel`, they probably introduced new checks.
Someone need to investigate and fix the code. If it's impossible to fix it w/o breaking something, we should add entries to https://github.com/ansible-collections/community.libvirt/blob/main/tests/sanity/ignore-2.20.txt to make CI ignore the violations.
If you want to fix it, it's vacant.

PR #215, opened 2025-09-03
Add an example of how to create a volume from a template

pablomh:
##### SUMMARY
Add an example of how to create a volume from a template

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
virt_volume

##### ADDITIONAL INFORMATION
The current examples only showed how to do it providing the xml content inline, so I think that using a template is more Ansible-like.


Andersson007:
@pablomh hello, thanks for the contribution!

PR #224, opened 2025-11-04
Add new force_disk parameter for cloud_virtual_instance module

thinkdoggie:
##### SUMMARY

This PR adds a new force_disk parameter to force deletion and recreation of existing disk files for virt_cloud_instance module. It could be useful to repetitive execution after a failed task.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME

`virt_cloud_instance`

##### ADDITIONAL INFORMATION

Task snippet in the test playbook
```yaml
  - name: Create OpenStack controller virtual machine
    community.libvirt.virt_cloud_instance:
      name: '{{ item }}'
      state: present
      base_image: '{{ openstack_node_base_image }}'
      image_cache_dir: '{{ cloudimg_cache_dir }}'
      image_checksum: '{{ openstack_node_base_image_checksum }}'
      force_pull: false
      force_disk: true
      url_timeout: 600
      memory: 32768
      memory_opts:
        current_memory: 16384
      vcpus: 4
      osinfo:
        name: '{{ openstack_node_osname }}'
      graphics:
        type: vnc
      iothreads: 1
      disks:
        - path: /srv/kvm/images/{{ item }}.qcow2
          size: 100
          bus: virtio
          cache: none
          format: qcow2
          driver:
            io: native
            discard: unmap
            detect_zeroes: unmap
      networks:
        - bridge: br0
          model:
            type: virtio
        - type: direct
          source: bond0
          source_opts:
            mode: bridge
          model:
            type: virtio
      cloud_init:
        meta_data: |
          local-hostname: {{ item }}
        user_data: |
          #cloud-config
          users:
            - name: sysadm
              sudo: ALL=(ALL) NOPASSWD:ALL
              shell: /bin/bash
              ssh_authorized_keys:
                - "{{ openstack_node_ssh_key }}"
          ssh_pwauth: false
          apt:
            preserve_sources_list: false
            primary:
              - arches: [default]
                uri: http://mirrors.ustc.edu.cn/ubuntu/
            security:
              - arches: [default]
                uri: http://mirrors.ustc.edu.cn/ubuntu/
        network_config: |
          version: 2
          ethernets:
            enp1s0:
              dhcp4: false
              addresses:
              - {{ hostvars[item].ansible_host }}/24
              routes:
              - to: default
                via: 192.168.10.254
              nameservers:
                addresses:
                - 192.168.10.1
                - 192.168.100.1
      wait_for_cloud_init_reboot: true
    loop: "{{ groups['openstack_controllers'] }}"
```

### DIFF
--- a/plugins/modules/virt_cloud_instance.py
+++ b/plugins/modules/virt_cloud_instance.py
@@ -56,6 +56,15 @@ options:
       - Only used when O(base_image) is a URL.
       - Useful for updating cached images to their latest versions.
     default: false
+  force_disk:
+    type: bool
+    description:
+      - Force deletion and recreation of existing disk files.
+      - When set to V(true), any existing disk files specified in O(disks) will be removed and recreated.
+      - When set to V(false), the module will fail if any disk file already exists.
+      - This parameter only affects disk file creation, not VM definition.
+      - Use with O(recreate=true) to recreate both the VM and its disk files.
+    default: false
   image_checksum:
     type: str
     description:
@@ -245,6 +254,21 @@ EXAMPLES = """
       meta_data:
         instance-id: web-server-01
         local-hostname: web-server-01
+
+# Force overwrite existing disk files
+- name: Create VM and force overwrite existing disk
+  community.libvirt.virt_cloud_instance:
+    name: test-vm
+    base_image: /srv/images/ubuntu-22.04-server-cloudimg-amd64.img
+    force_disk: true
+    disks:
+      - path: /var/lib/libvirt/images/test-vm.qcow2
+        size: 20
+        format: qcow2
+    memory: 2048
+    vcpus: 2
+    networks:
+      - network: default
 """
 
 RETURN = r"""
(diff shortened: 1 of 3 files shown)


plugins/modules/virt_cloud_instance.py
```diff
@@ -56,6 +56,15 @@
...
     default: false
+  force_disk:
+    type: bool
+    description:
+      - Force deletion and recreation of existing disk files.
+      - When set to V(true), any existing disk files specified in O(disks) will be removed and recreated.
+      - When set to V(false), the module will fail if any disk file already exists.
+      - This parameter only affects disk file creation, not VM definition.
+      - Use with O(recreate=true) to recreate both the VM and its disk files.
+    default: false
```
Andersson007:
```suggestion
    default: false
    version_added: '2.1.0'
```

Andersson007 (commented):
@thinkdoggie thanks for the feature! one minor thing from me. The rest LGTM

Is it not possible to cover it with integration tests?

plugins/modules/virt_cloud_instance.py
thinkdoggie: Since `virt_cloud_instance` is a unreleased module, it seems that there's no need to add the `version_added` property here.

Andersson007 (approved):

Andersson007:
@csmart could you please take a look? ideally before releasing 2.1.0

thinkdoggie:
@csmart Thanks for the thoughtful review! I understand your concerns about idempotency and safety. Let me address each point:

**On Idempotency:**
The `force_disk` parameter actually maintains idempotency - it just changes the *intended state*. When a user sets `force_disk=true`, they're explicitly declaring: "I want this disk file to be freshly created from the base image." Running the task multiple times with `force_disk=true` produces the same result: a fresh disk. This is similar to how ansible.builtin.copy` with `force=true` works. Especially, this would be useful to rerunnin after a failed task cleanup.

**Safe by Default:**
The default behavior remains safe and idempotent:
- `force_disk=false` (default): Module **fails** if disk exists, requiring manual cleanup
- `force_disk=true`: User explicitly opts-in to destructive behavior

This is similar to how `recreate=true` works for VM definitions - it's an explicit opt-in for destructive operations.

**On Shared Disks:**
This is a valid concern. However, if users are pointing multiple VMs at the same disk path, that's already a configuration error.

**Comparison with Other Collections:**
Looking at similar patterns:
- `community.docker.docker_container`: Has `recreate` parameter for similar behavior
- `community.general.terraform`: Has `force_init` for forced reinitialization
- Many modules use `force_*` parameters for explicit destructive actions

**Regarding Separate Tasks:**
While a separate cleanup task is certainly an option, having `force_disk` provides convenience for:
1. Atomic operations in testing scenarios
2. Avoiding race conditions between cleanup and creation
3. Simpler playbooks for development workflows

Issue #222, opened 2025-10-25
virt_install: cloud_init.user_data parameter documentation missing and format unclear


ksrp1984:
##### SUMMARY
The documentation for `community.libvirt.virt_install` module lacks clear examples and specifications for the `cloud_init.user_data` parameter. Currently, users face confusion about the expected format:

- When passing raw string data, the module expects a dictionary
- When passing dictionary data, the parameter formatting causes command-line errors
- No working examples are provided for cloud-init configuration

The documentation should clearly specify:
- The exact structure expected for `user_data` 
- Working examples for common cloud-init use cases
- How to properly format YAML vs dictionary input

##### ISSUE TYPE
- Documentation Report

##### COMPONENT NAME
`community.libvirt.virt_install` module documentation

##### ANSIBLE VERSION
ansible --version
ansible [core 2.16.3]
  config file = None
  configured module search path = ['/home/sergei/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3/dist-packages/ansible
  ansible collection location = /home/sergei/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/bin/ansible
  python version = 3.12.3 (main, Aug 14 2025, 17:47:21) [GCC 13.3.0] (/usr/bin/python3)
  jinja version = 3.1.2
  libyaml = True

thinkdoggie:
> Documentation Issue: Missing examples for PCI passthrough network configuration
> 
> The virt_install module fails with error "Unknown parameters --network: ['hostdev']" when trying to use PCI passthrough for network devices.
> 
> Expected behavior: The module should support hostdev parameter in networks section for PCI passthrough, as documented in libvirt.
> 
> Current behavior: Module only accepts basic network types (bridge, network) but not hostdev for direct device assignment.

Hi @ksrp1984 , I'll do an end-to-end test for PCI passthrough later.

Andersson007:
@thinkdoggie I've asked @csmart about our release plans, will let you know, thanks for pinging!

thinkdoggie:
> Documentation Issue: Missing examples for PCI passthrough network configuration
> 
> The virt_install module fails with error "Unknown parameters --network: ['hostdev']" when trying to use PCI passthrough for network devices.
> 
> Expected behavior: The module should support hostdev parameter in networks section for PCI passthrough, as documented in libvirt.
> 
[...]

PR #221, opened 2025-10-17
Enhance networks.source parameter for virt_install and virt_cloud_instance modules

thinkdoggie:
##### SUMMARY

This PR enhances the networks.source and networks.source_opts parameters for virt_install and virt_cloud_instance modules, including:
- Add new networks.source_opts parameter to provide additional macvtap options.
- Enhance networks.source parameter to support both string (macvtap host interface) and dictionary inputs.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME

`virt_install`
`virt_cloud_instance`

##### ADDITIONAL INFORMATION

Task snippet in the test playbook
```yaml
  - name: Create OpenStack controller virtual machine
    community.libvirt.virt_cloud_instance:
      name: '{{ item }}'
      state: present
      base_image: '{{ openstack_node_base_image }}'
      image_cache_dir: '{{ cloudimg_cache_dir }}'
      image_checksum: '{{ openstack_node_base_image_checksum }}'
      force_pull: false
      force_disk: true
      url_timeout: 600
      memory: 32768
      memory_opts:
        current_memory: 16384
      vcpus: 4
      osinfo:
        name: '{{ openstack_node_osname }}'
      graphics:
        type: vnc
      iothreads: 1
      disks:
        - path: /srv/kvm/images/{{ item }}.qcow2
          size: 100
          bus: virtio
          cache: none
          format: qcow2
          driver:
            io: native
            discard: unmap
            detect_zeroes: unmap
      networks:
        - bridge: br0
          model:
            type: virtio
        - type: direct
          source: bond0
          source_opts:
            mode: bridge
          model:
            type: virtio
      cloud_init:
        meta_data: |
          local-hostname: {{ item }}
        user_data: |
          #cloud-config
          users:
            - name: sysadm
              sudo: ALL=(ALL) NOPASSWD:ALL
              shell: /bin/bash
              ssh_authorized_keys:
                - "{{ openstack_node_ssh_key }}"
          ssh_pwauth: false
          apt:
            preserve_sources_list: false
            primary:
              - arches: [default]
                uri: http://mirrors.ustc.edu.cn/ubuntu/
            security:
              - arches: [default]
                uri: http://mirrors.ustc.edu.cn/ubuntu/
        network_config: |
          version: 2
          ethernets:
            enp1s0:
              dhcp4: false
              addresses:
              - {{ hostvars[item].ansible_host }}/24
              routes:
              - to: default
                via: 192.168.10.254
              nameservers:
                addresses:
[...]

PR #225, opened 2025-11-09
Enhance virt_install and virt_cloud_instance to support primary value attribute

thinkdoggie:
##### SUMMARY

This PR enhances the virt_install and virt_cloud_instance modules by adding support for a special primary value attribute (_value or value) in command line options. This feature is essential for properly handling virt-install device options where a primary value must be specified alongside additional attributes.

Many virt-install device options (like --hostdev, --watchdog, --serial, etc.) follow the pattern --option primary_value,attr1=val1,attr2=val2. Without the ability to specify a primary value, these options cannot be fully configured. The hostdev issue mentioned in #222 can be resolved by this PR.

The _dict2options function has been updated to prioritize _value over value and handle formatting correctly. This enhancement also allows simplification of the networks.source_opts parameter that was added in a previous PR (#221), as it can now be handled more elegantly using the primary value attribute. For example, `--network source=bond0,source.mode=bridge` can be defined as "source: {value: bond0, mode: bridge}" with the primary value for the nested `networks.source` dict. So, the previous PR #221 can be reverted safely before creating releaes 2.1.0. 

##### ISSUE TYPE
- Bugfix Pull Request
- Feature Pull Request

##### COMPONENT NAME
`virt_install`
`virt_cloud_instance`

##### ADDITIONAL INFORMATION

Task snippet in the test playbook.
```yaml
  - name: Create OpenStack controller virtual machine
    community.libvirt.virt_cloud_instance:
      name: '{{ item }}'
      state: present
      base_image: '{{ openstack_node_base_image }}'
      image_cache_dir: '{{ cloudimg_cache_dir }}'
      image_checksum: '{{ openstack_node_base_image_checksum }}'
      force_pull: false
      force_disk: true
      url_timeout: 600
      memory: 32768
      memory_opts:
        current_memory: 16384
      vcpus: 4
      osinfo:
        name: '{{ openstack_node_osname }}'
      graphics:
        type: vnc
      iothreads: 1
      disks:
        - path: /srv/kvm/images/{{ item }}.qcow2
          size: 100
          bus: virtio
          cache: none
          format: qcow2
          driver:
            io: native
            discard: unmap
            detect_zeroes: unmap
      networks:
        - bridge: br0
          model:
            type: virtio
      hostdev:
        value: "81:00.0"
      cloud_init:
        meta_data: |
          local-hostname: {{ item }}
        user_data: |
          #cloud-config
          users:
            - name: sysadm
              sudo: ALL=(ALL) NOPASSWD:ALL
              shell: /bin/bash
              ssh_authorized_keys:
                - "{{ openstack_node_ssh_key }}"
          ssh_pwauth: false
          apt:
            preserve_sources_list: false
            primary:
              - arches: [default]
                uri: http://mirrors.ustc.edu.cn/ubuntu/
            security:
              - arches: [default]
                uri: http://mirrors.ustc.edu.cn/ubuntu/
        network_config: |
          version: 2
          ethernets:
            enp1s0:
              dhcp4: false
              addresses:
              - {{ hostvars[item].ansible_host }}/24
              routes:
              - to: default
                via: 192.168.10.254
              nameservers:
                addresses:
                - 192.168.10.1
                - 192.168.100.1
      wait_for_cloud_init_reboot: true
    loop: "{{ groups['openstack_controllers'] }}"
```

### DIFF
--- a/changelogs/fragments/221-networks-source-opts.yaml
+++ b/changelogs/fragments/221-networks-source-opts.yaml
@@ -1,5 +1,0 @@
-minor_changes:
-  - virt_install - added new networks.source_opts parameter to provide additional macvtap options.
-  - virt_cloud_instance - added new networks.source_opts parameter to provide additional macvtap options.
-  - virt_install - enhanced networks.source parameter to support both string and dictionary inputs.
-  - virt_cloud_instance - enhanced networks.source parameter to support both string and dictionary inputs.
--- a/changelogs/fragments/225-primary-value-support.yml
+++ b/changelogs/fragments/225-primary-value-support.yml
@@ -1,0 +1,3 @@
+minor_changes:
+  - virt_install - added support for primary value attribute (_value or value) in dynamic dict options that require a primary value alongside additional attributes.
+  - virt_cloud_instance - added support for primary value attribute (_value or value) in dynamic dict options that require a primary value alongside additional attributes.
--- a/plugins/doc_fragments/virt_install.py
+++ b/plugins/doc_fragments/virt_install.py
@@ -1301,19 +1301,12 @@ options:
           - Specify the interface ROM BIOS configuration
           - The dictionary contains key/value pairs that define individual properties.
       source:
-        type: raw
-        description:
-          - Specify the source host interface.
-          - Can be either a string of IFACE name or a dictionary with detailed configuration.
-          - When provided as a string, use I(source_opts) to specify additional source properties.
-          - "Example as string: V(bond0)"
-          - "Example with source_opts: I(source=bond0) and I(source_opts={mode: bridge})"
-      source_opts:
         type: dict
         description:
-          - Additional options for the direct attached macvtap interface.
+          - Specify the details of the direct attached macvtap interface.
           - The dictionary contains key/value pairs that define individual properties.
-        version_added: '2.1.0'
+          - "You can use the special attribute I(value) to specify a primary value that appears before other properties in the command line."
+          - "Example: I(source={value: bond0, mode: bridge})"
       target:
         type: dict
         description:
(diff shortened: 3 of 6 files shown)


csmart:
I had just pushed a branch for `2.1.0`, so good timing :smile: So I think this change is [similar to my feedback](https://github.com/ansible-collections/community.libvirt/pull/221#issuecomment-3465311037) in #221?

I don't think we need to revert #221, you have made the necessary changes, we can just merge this as is.

Is it possible to include a clear example in the docs that makes use of these new features?

thinkdoggie:
> I had just pushed a branch for `2.1.0`, so good timing 😄 So I think this change is [similar to my feedback](https://github.com/ansible-collections/community.libvirt/pull/221#issuecomment-3465311037) in #221?
> 
> I don't think we need to revert #221, you have made the necessary changes, we can just merge this as is.
> 
> Is it possible to include a clear example in the docs that makes use of these new features?

Hi @csmart, correct. This change address your concern about the dict extension in #221. However, I didn't work out a good solution at that time.

If we don't revert #221, it means we have to maintain the backward compatibility for options `networks.source_opts`. If reverted, we can simply keep the type of  `networks.source` as dict rather than raw.

Futhermore, the hostdev and other configurations with device options won't work if the primary value feature is not supported. So, I hope we can include it in 2.1.0 release.

Thanks!

csmart:
But we never released a version with `source_opts` support I don't think, so what's the benefit of reverting #221 and then adding this new patch on top, instead of just having this new patch on top that removes it anyway?

I mean if you want to put in a MR that reverts #221 and then re-work this patch on top of it, feel free..

thinkdoggie:
In addition, an example is also added to the the example section of the virt_install module documentation.
```yaml
# VM with PCI passthrough for a network device
- name: Create VM with PCI network passthrough
    name: passthrough-vm
    memory: 4096
    vcpus: 4
    disks:
      - size: 30
    osinfo:
      name: ubuntu20.04
    import: true
    host_devices:
      - value: "81:00.0"
```

I also added an example for the `networks.source` option doc:
```yaml
          - "Example: I(source={value: bond0, mode: bridge})"
```

Andersson007:
@thinkdoggie @mwpastore @csmart thanks a lot for the contribution!

PR #227, opened 2025-11-23
Common utils and livbirt secret

arddennis:
##### SUMMARY

A new module virt_secret which allows to manage libvirt secrets and their value. Module has full support of check and diff mode and introduce new base module definition proposal, defined in module_utils/common.py. This common class is defines common behavior of Ansible module for this collection. A handy way to exit, form diff or add warn messages.

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
virt_secret

##### ADDITIONAL INFORMATION

New module is written in python 3.10 and is not compatible with version 1 of this collection. Recent python version was taken due to minimal python version supported by ansible-core==2.17 is 3.10. But even 2.16 requires 3.10-3.12

**module_utils/common.py** 
contains base class **VirtModule** with initialization of libvirt connection and typical exit, check_imports and abstract methods. Additionally there is a dataclass **ModuleStatus** which is used to collect existing status information. Separate class is needed to be able to fine control behavior of the status report. Which then can be accessed just as class property.

**libvirt_error_to_none** is example of the decorator which changes libvirterror exceptions into none output. Suitable when you try to get object from libvirt which does not exist. We should get None instead.

**compare_dicts** and **compare_lists** provided as recursive objects comparsion functions which does not require external libraries.

**data_as_dict** is used to produce normal dict from the dataclass object. Done recursively, so as result we get consumable by ansible dict.

**module/virt_secret.py**

This is actually new module. It can manage libvirt secrets. Illustrate how above common classes can be used to write a module with diff and check support.

All defined or received values converted to the dataclass **SecretElement** which can be compared and used to produce the diff. It has xml representation used by libvirt api similar to the other modules in the collection.

P.S. **SecretUsage** provided as exmaple, it can be used instead of str in appropriate filed of **SecretElement**

Tests are basic as I don't know if this module is suitable or not. Plus I might need help and guidance in this process.

### DIFF
--- a/plugins/modules/virt_secret.py
+++ b/plugins/modules/virt_secret.py
@@ -156,7 +156,7 @@ attributes:
 RETURN = r'''
 secret_xml:
   type: str
-  description: When I(command=get_xml) return xml definition of the secret
+  description: When I(command=get_xml) returns XML definition of the secret.
   returned: success
   sample: |
     <secret ephemeral="no" private="yes">
@@ -170,7 +170,7 @@ list_secrets:
   type: list
   elements: str
   returned: success
-  description: When (command=list_secrets) returns secrets_list of secrets in xml format
+  description: When (command=list_secrets) returns a list of secrets in XML format.
   sample:
     - |
       <secret ephemeral='no' private='yes'>
@@ -194,7 +194,7 @@ list_secrets:
 
 EXAMPLES = r'''
 ---
-- name: Create new secret using xml definition and command
+- name: Create new secret using xml option and command
   community.libvirt.virt_secret:
     command: create
     xml: |
@@ -483,7 +483,7 @@ class LibvirtSecretModule(VirtModule):
         """ Forms defined XML
 
         It looks like this function has chicken and the egg problem, but if
-        xml property is defined, defined_element will be calculated first based
+        xml option is defined, defined_element will be calculated first based
         on its values. If it is not, we expect secret to be defined and will
         use it instead.
         """
@@ -559,7 +559,7 @@ class LibvirtSecretModule(VirtModule):
     def _parse_xml_secret(self, xml: str) -> Union[SecretElement, None]:
         """ Parse xml string and return it as SecretElement object
 
-        :param xml: string containing xml of the secret. Param left for future
+        :param xml: string containing XML of the secret. Param left for future
                     reuse of the function.
         """
         an_xml = etree.fromstring(xml)
@@ -621,7 +621,7 @@ class LibvirtSecretModule(VirtModule):
         self.exit()
 
     def list_secrets(self):
-        """ Get list of defined secrets and return them as list of xmls """
+        """ Get list of defined secrets and return them as list of XMLs """
 
         current_secrets = [
             secret.XMLDesc()


arddennis:
> 2\. major_changes

OK, understood. This does not touch anything existing and does not need to be a major change. I will remove fragment. Plus both a new module approach and the module itself is not used anywhere else in the modules. This can be used in the future to change existing modules or write a new versions alongside. So does it still need to be split?

It is just a proposal and POC of approach. Module is fully functional and cover all libvirt secrets functions. So this is a good value to accept. It is not compatible with python 2, but it wasn't present in version 1. 

As for the concerns about backporing the changes. Maybe a set of the new modules can be placed alongside with the others and marked future. They can pass the same set of tests. I don't have good answer here. Just thought ansible does not support python 2 for some time already. And older versions aren't maintained/updated any longer.

P.S. I honestly don't know how to open a discussion and purpose a way for the modules to behave. Though writing a module will be a good example and starting point to get a feedback. Creation of Issue in the repo without fully done example felt strange.

Andersson007:
> IMO, backports don't _have to_ be clean cherry-picks to a stable branch, while it totally is a _nice to_ have. As the whole point of stable branch policy, is to be able to have breaking changes on `main` branch, but be able to patch stable versions through backport process, where backported patches may need adoption to work properly with older code. Otherwise you're de-facto actively developing version 1, if changes on main should not break it in any way.

@noonedeadpunk I also think they don't have to. IMO it's OK in case the changes are not very complex otherwise can make maintainers life hard. This part confuses me, especially "Use common class for the libvirt modules":
 ```
This is a proposal to switch to a python3 style. Use common class for the libvirt modules which supports diffs, check mode and can add be used to control ansible module during a runtime (able to exit on error/exception and produce explanation)
```
That's why i asked for clarifications and mentioned the caveats.

> It seems this PR specifically is adding a new module and does not touch anything existing. So it does satisfy this criteria?
It does not introduce any refactoring to any existing module from what I see.

If there's no refactoring and the PR contains only changes related to the new module, it can be merged here of course, no problem.

**Review by Andersson007 — COMMENTED** [review]
@arddennis here are some general things to fix. I can't spend any time on code analysis unfortunately. Please take a look

**Comment by Andersson007** [comment]
> Oh, ok, I agree this is confusing. I was checking the diff of the PR and did not read the intro carefully 😁

@noonedeadpunk , hehe, I in turn took a look at the description and the scary-looking-full-of-major-changes changelog fragment and didn't take a look at the code:) Decided to ask / mention caveats just in case.

File: plugins/modules/virt_secret.py [hunk 1]
```diff
@@ -0,0 +1,664 @@
...
+DOCUMENTATION = r'''
+---
+module: virt_secret
+version_added: '2.1.0'
+author:
+  - Denys Mishchenko (@arddennis)
+
+short_description: Manage libvirt secrets and their values
+description:
+  - Manage I(libvirt) secrets. Can add remove or update secrets in libvirt
```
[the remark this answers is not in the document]
arddennis: docstring was updated according to provided documentation.

plugins/modules/virt_secret.py
```diff
@@ -0,0 +1,664 @@
...
+    password: somesecureandrandomsecret1234
+
+- name: Remove vTPM secret
+  community.libvirt.virt_secret:
+    uuid: 57ea8fd0-9b82-4e54-9d16-df7d2765844d
+    state: absent
+'''
+
+RETURN = r'''
+secretXML:
```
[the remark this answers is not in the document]
arddennis: refactored return variable including integration tests. Now it secret_xml

plugins/modules/virt_secret.py
```diff
@@ -0,0 +1,664 @@
...
+      <uuid>e4b5978c-ba37-5605-97c1-4a20413d0fc9</uuid>
+      <description>test_secret pool secret</description>
+      <usage type="ceph">
+        <name>test_secret</name>
+      </usage>
+    </secret>"
+list_secrets:
+  type: list
+  returned: success
+  description: When I(command=list_secrets) returns secrets_list of secrets in xml format
```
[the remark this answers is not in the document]
arddennis: Corrected. Where O() formatting is used it is lowercase, when we talk about format itself, it is uppercase.

plugins/module_utils/common.py
```diff
@@ -0,0 +1,239 @@
...
+except ImportError as safe_dump_import_exception:
+    SAFE_DUMP_IMPORT_ERR = safe_dump_import_exception
+else:
+    SAFE_DUMP_IMPORT_ERR = None
+
+
+STATE_CHOICES = ['present', 'absent']
+
+
+def compare_dicts(dict_one: dict, dict_two: dict) -> bool:
```
csmart: I don't think we're doing anything special with values here, do we need to have a dedicated recursive function for this? Could python's built in `==` operator suffice? It should ignore order and support nested operations.

```
>>> def compare_dicts(dict_one: dict, dict_two: dict) -> bool:
...     return dict_one == dict_two

>>> compare_dicts({"a": 1, "b": 2}, {"b": 2, "a": 1})
True

>>> compare_dicts({"b": 2, "a": 1}, {"a": 1, "b": 2})
True

>>> compare_dicts({"b": 2, "a": 1}, {"a": 1})
False

>>> compare_dicts({"b": 2}, {"a": 1, "b": 2})
False

>>> compare_dicts({"b": 2, "a": 1}, {"a": 1, "b": {"this": "that"}})
False

>>> compare_dicts({"a": 1, "b": {"this": "that"}}, {"b": {"this": "that"}, "a": 1})
True

>>> compare_dicts({"b": {"this": "that"}, "a": 1}, {"a": 1, "b": {"this": "that"}})
True
```

plugins/module_utils/common.py
```diff
@@ -0,0 +1,239 @@
...
+        else:
+            dict_item = asdict(item)
+        result.append(dict_item)
+    return result
+
+
+def parse_xml_etree(xml: str) -> dict:
+    """ Parse xml etree and return it as dict """
+    result = {}
+    for element in etree.fromstring(xml).iterchildren():
```
csmart: is the input here always valid XML, or should it be wrapped in `try`? and do we need to deal with any nested elements?

plugins/modules/virt_secret.py
```diff
@@ -0,0 +1,718 @@
...
+          - vtpm
+      usage_id:
+        description:
+          - Defines unique secret_ID. Unique for each O(secret.usage).
+        type: str
+      description:
+        type: str
+        description:
+          - Defines secret description.
+  password:
```
csmart: will this be `no_log`?

plugins/modules/virt_secret.py
```diff
@@ -0,0 +1,718 @@
...
+  check_mode:
+    description: Check mode is fully supported.
+    support: full
+    details:
+      - In check mode, secrets are not actually created, updated or deleted.
+      - |
+        Module compares existing state of the secret in libvirt and check if
+        change is required.
+      - Will always produce changed with O(command=set_value).
+  diff_mode:
```
csmart: how should we deal with `no_log` with `diff`?

plugins/module_utils/common.py
csmart: if you agree, I think it's better to remove code that's not explicitly required for the change at hand

csmart:
> @csmart @arddennis any updates on this? are we happy to merge it or there's something that still needs to be done or disagreement?


@Andersson007 I think there were a few small things to fix up after my code review, as well as the conversation about whether to introduce new functions now that aren't yet used, or keep it focused on what's required for the libvirt secret component, which I prefer the latter.

PR #241, opened 2026-02-24
feat: add commands return value for provisioning tools

thinkdoggie:
Return the underlying qemu-img and virt-install commands executed during provisioning via a new `commands` field.

##### SUMMARY

Add a commands return field to the virt_install and virt_cloud_instance modules, exposing the exact CLI commands executed during VM provisioning.

- virt_install: returns the virt-install command used to create the VM
- virt_cloud_instance: returns the full sequence of commands — qemu-img operations for disk preparation followed by the virt-install command

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME
`virt_install`
`virt_cloud_instance`

##### ADDITIONAL INFORMATION
N/A

### DIFF
--- a/plugins/modules/virt_install.py
+++ b/plugins/modules/virt_install.py
@@ -320,7 +320,18 @@ EXAMPLES = """
     state: absent
 """
 
-RETURN = r""" # """
+RETURN = r"""
+commands:
+    description:
+        - List of virt-install commands executed by this module.
+        - Only populated when a VM is created or recreated (V(state=present) and VM did not already exist, or V(recreate=true)).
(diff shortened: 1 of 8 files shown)


plugins/modules/virt_install.py
```diff
@@ -340,8 +351,9 @@ def core(module):
 
     result = dict(
         changed=False,
-        orignal_message="",
+        original_message="",
```
Andersson007: @thinkdoggie this is a breaking change, should be released in a major release. Moreover, it unrelated to this particular feature, could we remove it?

PR #246, opened 2026-03-24
Add configuration option to disable alternate id group creation

Cabub:
##### SUMMARY

This change adds a new boolean configuration option `alternate_id_groups` to the inventory plugin. Setting it to `false` means the discovered hosts will no longer be added to auto-generated groups based on the alternate id (name or UUID). The default setting is `true`, preserving existing functionality.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
`plugins/inventory/libvirt.py`

##### ADDITIONAL INFORMATION

EOT:

```sh
$ ansible-test units --color -v --docker default
<redacted for brevity>
============================= 304 passed in 3.59s ==============================
```

And on my homelab, with this config:

```sh
plugin: community.libvirt.libvirt
uri: 'qemu+ssh://cabub@lab/system'

inventory_hostname: name
alternate_id_groups: false
```

The resulting hosts are not assigned to uuid groups:

```sh
ansible-inventory -i inventory --graph       
libvirt: QEMU Driver error : Requested operation is not valid: domain is not running
libvirt: Domain Config error : Requested operation is not valid: domain is not running
libvirt: QEMU Driver error : Requested operation is not valid: domain is not running
libvirt: Domain Config error : Requested operation is not valid: domain is not running
libvirt: QEMU Driver error : Requested operation is not valid: domain is not running
libvirt: Domain Config error : Requested operation is not valid: domain is not running
@all:
  |--@ungrouped:
  |  |--router
  |  |--data
  |  |--bastion
```

### DIFF
--- a/changelogs/fragments/246-add-alias-groups-option.yml
+++ b/changelogs/fragments/246-add-alias-groups-option.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - inventory - added option ``alternate_id_groups`` (default ``true``)
--- a/plugins/inventory/libvirt.py
+++ b/plugins/inventory/libvirt.py
@@ -25,8 +25,7 @@ options:
     inventory_hostname:
         description: |
             What to register as the inventory hostname.
-            If set to 'uuid' the uuid of the server will be used and a
-            group will be created for the server name.
+            If set to 'uuid' the uuid of the server will be used.
             If set to 'name' the name of the server will be used unless
             there are more than one server with the same name in which
             case the 'uuid' logic will be used.
@@ -36,6 +35,12 @@ options:
             - name
             - uuid
         default: "name"
+    alternate_id_groups:
+        description: |
+            Determines whether groups will be created using the alternate id
+            (hostname or UUID)
+        type: bool
+        default: true
 '''
 
 EXAMPLES = r'''
@@ -114,8 +119,10 @@ class InventoryModule(BaseInventoryPlugin, Constructable):
             # TODO(daveol): Fix "Invalid characters were found in group names"
             # This warning is generated because of uuid's
             self.inventory.add_host(inventory_hostname)
-            self.inventory.add_group(inventory_hostname_alias)
-            self.inventory.add_child(inventory_hostname_alias, inventory_hostname)
+
+            if self.get_option('alternate_id_groups'):
+                self.inventory.add_group(inventory_hostname_alias)
+                self.inventory.add_child(inventory_hostname_alias, inventory_hostname)
 
             if connection_plugin is not None:
                 self.inventory.set_variable(


thinkdoggie:
@Cabub,

Thanks for working on this. Adding a toggle to disable the auto-generated alias/UUID (or name) grouping is a meaningful and user-friendly improvement, and keeping the default behavior unchanged preserves backwards compatibility.

#### Requested follow-ups before merge

- Add a changelog fragment: This introduces a user-facing configuration option, so please add an entry under changelogs/fragments/ (likely minor_changes) describing the new toggle and its effect.
- Update `inventory_hostname` documentation: The current inventory_hostname description states that a group “will be created” for the alternate identifier. With this change it becomes conditional, so please update the docs to clarify that group creation is controlled by the new option.

#### Naming suggestion

Consider the option name: `create_alias_groups` matches the internal implementation, but from a user perspective “alias” is a new, undefined concept and may be confusing.

Would you consider renaming it to something more self-explanatory, e.g. `alternate_id_groups`, to make it immediately clear that it refers to groups based on the alternate identifier (name ↔ UUID)?

If you prefer to keep `create_alias_groups`, I’d recommend making the option description explicitly define what “alias” means in this context (alternate identifier derived from the inventory_hostname choice).

Cabub:
Great suggestions! Thanks for the input. I will add the changelog fragment and update the option name. After I initially opened the PR, I stumbled upon the [offical documentation](https://docs.ansible.com/projects/ansible/devel/community/create_pr_quick_start.html) regarding setting up a dev environment, and have been working my way through it in order to comply with the conventions of the project.

I've implemented your suggestions, and I took a look at the tests to see if I could spot a way to add tests here, but I didn't see one. Given that this change is basically just a conditional that relies entirely on upstream ansible functionality, I'm not sure it's totally necessary to add a test here.

thinkdoggie:
Thanks for the updates.

On testing: I agree it’s not a hard requirement for this PR. Inventory plugins are difficult to integration-test reliably in CI because they depend on a working libvirt environment, and this change is essentially a small conditional with the default preserving existing behavior. I’m fine merging without adding new tests here.

That said, it would still be valuable to improve coverage later. We can add unit tests with mocked libvirt to validate host/group generation (including alternate_id_groups: false) in a stable way.

If you’ve already run ansible-test sanity and ansible-test units locally, we should be good to proceed. I also saw that all CI checks passed.

thinkdoggie (approved):

thinkdoggie:
@Cabub 
In addition, could you add some snippet or test output under section `ADDITIONAL INFORMATION` in the description?

thinkdoggie:
@Cabub could you also squash all commits into single one in this PR? Thanks!

Cabub:
I noticed a failing sanity check in the CI, but it's not from something I changed:

```
00:15 ERROR: Found 1 pylint issue(s) which need to be resolved:
00:15 ERROR: plugins/inventory/libvirt.py:58:0: ansible-bad-import-from: Import raise_from from the Python standard library equivalent instead of ansible.module_utils.six (50%)
```

The LOC in libvirt.py:

```python
from ansible.module_utils.six import raise_from
```

thinkdoggie:
@Cabub I merged #247. Could confirm if this PR should be rebased from main to merge `plugins/inventory/libvirt.py`? Thanks!

Cabub:
:heavy_check_mark: Done! Thanks for the merges!

PR #257, opened 2026-04-25
inventory: add filter option to select vms by regex

csmart:
Add a filter option (default `.*`) to select only specific VMs by running regex on name or UUID. This uses `re.search` which matches anywhere in the string, combine with `^` and `$` for an exact match.

Replacement for #101

### DIFF
--- a/plugins/inventory/libvirt.py
+++ b/plugins/inventory/libvirt.py
@@ -41,6 +41,14 @@ options:
             (hostname or UUID)
         type: bool
         default: true
+    filter:
+        description: |
+            Regex applied to each domain's name or UUID (whichever
+            ``inventory_hostname`` selects). Domains are skipped unless
+            the regex matches. Uses Python ``re.search`` semantics, so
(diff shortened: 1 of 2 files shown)


plugins/inventory/libvirt.py
```diff
@@ -41,6 +41,14 @@
...
         type: bool
         default: true
+    filter:
+        description: |
+            Regex applied to each domain's name or UUID (whichever
+            ``inventory_hostname`` selects). Domains are skipped unless
+            the regex matches. Uses Python ``re.search`` semantics, so
+            anchor with ``^`` / ``$`` for an exact match.
+        type: str
+        default: ".*"
```
Andersson007, line 51:
```suggestion
        default: ".*"
        version_added: '2.3.0'
```

Andersson007 (commented):
@csmart thanks for the PR!

cc @thinkdoggie

Andersson007 (approved):

thinkdoggie (approved):

PR #258, opened 2026-04-25
ci: update devel and add 2.21

csmart:
Devel is updated to 2.22.0 which bumps Fedora to 44, add 2.21.

* https://forum.ansible.com/t/ansible-core-devel-and-milestone-bumped-to-2-22-0-dev0/45646
* https://forum.ansible.com/t/ansible-test-devel-fedora-43-replaced-with-fedora-44/45695/


csmart:
@thinkdoggie I will merge this and then rebase #257

PR #261, opened 2026-05-01
inventory: add filter option to select vms by regex (#257)

csmart:
* inventory: add filter option to select vms by regex

Add a filter option (default `.*`) to select only specific VMs by running regex on name or UUID. This uses `re.search` which matches anywhere in the string, combine with `^` and `$` for an exact match.

I made such a mess of this because I used the UI.. sorry.


csmart:
this is crazy, I had no idea GH would do it like this - but @yorick1989 is on the author list now for this commit, thanks again @yorick1989 !

yorick1989:
Thank you for getting it to work and the merge!

Grtz,
Yorick

Issue #124, opened 2022-05-24
Add the ability to get the interface addresses of a Domain (VM)


jessecooper:
##### SUMMARY
I do not see this implemented, documented or discussed anywhere so, I am created this feature request with the good intention of implementing it if the feature is not already available.

I would like to be able to get the Interfaces Addresses of a Domain (VM).
  
virsh command:
`virsh domifaddr --source <agent|arp|lease> <vm_name>`
##### ISSUE TYPE
- The idea of this feature would comes from wanting to get the IP address of an interface of a VM that has just been deployed to continue the VMs OS setup. 

##### COMPONENT NAME
It seems the best place for this feature would be in the `virt` module.  The `VM_COMMAND` could be `get_ifaddresses`.

##### ADDITIONAL INFORMATION

This module would be used to get the Domain (VM) interface addresses.
Example code using [libvirt-python](https://libvirt.org/docs/libvirt-appdev-guide-python/en-US/html/libvirt_application_development_guide_using_python-Network_Interface-Addresses.html) 

```yaml
- name: vm info
   community.libvirt.virt:
     command: get_ifaddresses
     source: <agent|arp|lease>
     domain: <name_of_vm>
   register: ifaddresses
```

thinkdoggie:
Fixed by #266

PR #266, opened 2026-05-27
feat(modules/virt): add get_ifaddresses function to retrive domain in…

mark-starikov-lynxx:
##### SUMMARY
Adding get_interfaces function described in #124 

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
module_utils/libvirt.py
modules/virt.py

##### ADDITIONAL INFORMATION
using in tasks, flags can be either lease, arp or agent. e.g. 
```yaml
- name: get vm
  community.libvirt.virt:
    name: "opensuse_tumbleweed_latest"
    command: get_ifaddresses
    flags: arp
  register: vm_info
- name: print vm info
  ansible.builtin.debug:
    msg: "{{ vm_info }}"
```
will return something like
```json
ok: [localhost] => {
    "msg": {
        "changed": false,
        "failed": false,
        "vnet33": {
            "addrs": [
                {
                    "addr": "192.168.122.165",
                    "prefix": 0,
                    "type": 0
                }
            ],
            "hwaddr": "52:54:00:e7:85:6f"
        }
    }
}
```

### DIFF
--- a/plugins/module_utils/libvirt.py
+++ b/plugins/module_utils/libvirt.py
@@ -35,11 +35,17 @@ VIRT_STATE_NAME_MAP = {
     5: 'shutdown',
     6: 'crashed',
 }
-
+INTERFACE_ADDRESS_SOURCE_MAP = {
+    'lease': libvirt.VIR_DOMAIN_INTERFACE_ADDRESSES_SRC_LEASE,
+    'agent': libvirt.VIR_DOMAIN_INTERFACE_ADDRESSES_SRC_AGENT,
+    'arp': libvirt.VIR_DOMAIN_INTERFACE_ADDRESSES_SRC_ARP
+}
 
 class VMNotFound(Exception):
     pass
 
+class InvalidAddressSourceType(Exception):
+    pass
 
 class LibvirtConnection(object):
 
@@ -53,8 +59,7 @@ class LibvirtConnection(object):
         if "xen" in stdout:
             conn = libvirt.open(None)
         elif "esx" in uri:
-            auth = [[libvirt.VIR_CRED_AUTHNAME,
-                     libvirt.VIR_CRED_NOECHOPROMPT], [], None]
+            auth = [[libvirt.VIR_CRED_AUTHNAME,libvirt.VIR_CRED_NOECHOPROMPT], [], None]
             conn = libvirt.openAuth(uri, auth)
         else:
             conn = libvirt.open(uri)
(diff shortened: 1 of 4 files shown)


thinkdoggie:
The code looks good to me. But the CI pipeline was broken due to some sanity tests. 

@mark-starikov-lynxx , could you take a look at this? Probably the failures are not caused by your code changes. Or I will do further investigation later.

thinkdoggie:
Thanks for the contribution! I checked the CI failure and found the root cause.

`validate-modules` imports modules to introspect `argument_spec`. In this PR, `plugins/module_utils/libvirt.py` defines `INTERFACE_ADDRESS_SOURCE_MAP` at import time and directly references `libvirt.VIR_DOMAIN_INTERFACE_ADDRESSES_SRC_*`.

In sanity environments where `python-libvirt` is not installed, `import libvirt` fails, so `libvirt` is undefined. The new top-level map then raises `NameError`, which causes the `import-error` failures you see.  
This is why multiple modules fail together (`virt.py`, `virt_cloud_instance.py`, `virt_install.py`, `virt_secret.py`) — they all depend on this shared module util.

### Suggested fix
Avoid resolving `libvirt` constants at module import time. Build the source map lazily inside `get_ifaddresses()` (or guard it with `HAS_VIRT`).

After that, rerun sanity (especially `validate-modules`) to verify.

plugins/module_utils/libvirt.py
```diff
@@ -35,11 +35,17 @@
     5: 'shutdown',
     6: 'crashed',
 }
-
+INTERFACE_ADDRESS_SOURCE_MAP = {
```
thinkdoggie: Avoid resolving libvirt constants at module import time. Build the source map lazily inside `get_ifaddresses()` (or guard it with `HAS_VIRT`).

plugins/module_utils/libvirt.py
```diff
@@ -151,6 +156,14 @@ def get_uuid(self, vmid):
         vm = self.conn.lookupByName(vmid)
         return vm.UUIDString()
 
+    def get_ifaddresses(self, vmid, flag):
```
thinkdoggie, line 157: **Suggested Fix:**

```python
def get_ifaddresses(self, vmid, flag):
    if not HAS_VIRT:
        raise Exception("python-libvirt is required for get_ifaddresses")

    interface_address_source_map = {
        'lease': libvirt.VIR_DOMAIN_INTERFACE_ADDRESSES_SRC_LEASE,
        'agent': libvirt.VIR_DOMAIN_INTERFACE_ADDRESSES_SRC_AGENT,
        'arp': libvirt.VIR_DOMAIN_INTERFACE_ADDRESSES_SRC_ARP,
    }

   ...
```

mark-starikov-lynxx: thanks for suggested fix @thinkdoggie. I've put that in and also fixed some of the PEP8 errors from CI sanity checks. It looks mostly green now but please let me know if there are any other issues.

thinkdoggie (approved):

thinkdoggie:
@mark-starikov-lynxx Looks good to me. Thanks for your contribution. I'll merge this PR soon.

PR #268, opened 2026-06-05
fix: make install.os optional to support no_install without OS (#250)

thinkdoggie:
##### SUMMARY

The install.os parameter was incorrectly marked as required=True in the argument spec, preventing install.no_install=true from working without specifying an OS. When no_install is set, virt-install just creates the VM without an install phase, so OS is unnecessary.

Fixes #250 

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
`virt_install`

### DIFF
--- a/changelogs/fragments/268-fix-no-install-os-required.yaml
+++ b/changelogs/fragments/268-fix-no-install-os-required.yaml
@@ -1,0 +1,2 @@
+bugfixes:
+  - "virt_install - remove incorrect ``required=True`` from ``install.os`` parameter to allow ``install.no_install=true`` without specifying an OS."
--- a/plugins/doc_fragments/virt_install.py
+++ b/plugins/doc_fragments/virt_install.py
@@ -842,7 +842,7 @@ options:
         type: str
         description:
           - The OS name from I(libosinfo), e.g. V(fedora29).
-        required: true
+          - This is not required when O(install.no_install=true) is set.
       kernel:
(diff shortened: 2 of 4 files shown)


thinkdoggie:
@kristvanbesien, could you help review this PR? Thanks!

thinkdoggie:
@csmart @Andersson007 could you review this PR? Thanks!

csmart:
@thinkdoggie thanks for the patch. The fix looks correct to me, `install.os` shouldn't have been unconditionally required when using `no_install` :+1: 

The tests only validate the argspec change, it would be good if we can introduce a behavioural test that builds the command and validate it - not a blocker for this PR, but worth considering as a follow-up.

Speaking follow-ups, I tested the equivalent `virt-install` commands on fedora 44 and recent versions of `virt-install` now hard-require an OS name even with `--import` / `no_install` (with a fallback to `generic` if you force it off). So while this PR correctly removes the incorrect Ansible restriction, users will still usually want to provide an OS (via `install.os` or the top-level `osinfo` parameter) for good defaults. This is probably something we need to think about and I expect we will get more friction with `virt-install` v4+ as it becomes stricter around osinfo, defaults, and option validation...

thinkdoggie:
@csmart  Thanks for the review and for merging. Good point on the behavioral test, I'll follow up with a unit test for command building.

PR #272, opened 2026-06-18
feat(virt_install): add wait_timeout parameter for multi-phase installs

### LINKED ISSUE #253: events not set correctly
##### SUMMARY

The settings for events are not respected.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
community.libvirt.virt_install

##### ANSIBLE VERSION

```
ansible [core 2.20.4]
  config file = /Users/kvanbesi/.ansible.cfg
  configured module search path = ['/Users/kvanbesi/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /Users/kvanbesi/.local/pipx/venvs/ansible/lib/python3.14/site-packages/ansible
  ansible collection location = /Users/kvanbesi/.ansible/collections:/usr/share/ansible/collections
  executable location = /Users/kvanbesi/.local/bin/ansible
  python version = 3.14.3 (main, Feb  3 2026, 15:32:20) [Clang 17.0.0 (clang-1700.6.3.2)] (/Users/kvanbesi/.local/pipx/venvs/ansible/bin/python)
  jinja version = 3.1.6
  pyyaml version = 6.0.3 (with libyaml v0.2.5)
```

##### COLLECTION VERSION
-->
```
Collection        Version
----------------- -------
community.libvirt 2.2.1
```

##### OS / ENVIRONMENT

MacOS 26.3.1
Ansible has been installed with pipx
Target is a Centos10 system with libvirt.

##### STEPS TO REPRODUCE
```yaml
    - name: Install the VMs
        name: "{{ vm.name }}"
        autostart: true
        hvm: true
        transient: false
        memory: "{{ vm.ram * 1024 }}"
        vcpus: "{{ vm.cpu  }}"
        osinfo:
          name: rhel9.7
        events:
          on_poweroff: destroy
          on_reboot: restart
          on_crash: preserve
        disks:
          - size: "{{ vm.root_disk }}"
          - "{{  omit if vm.data_disk is not defined else {'size': vm.data_disk} }}"
        networks:
          - bridge: "{{ vm.net }}"
            mac:
              address: "{{ vm.mac }}"
        cdrom: /var/lib/libvirt/images/agent.x86_64.iso
        graphics:
          type: vnc
      loop: "{{ libvirt_vms }}"
      loop_control:
        loop_var: vm
```

##### EXPECTED RESULTS

I expected the VM to be created with:

  <on_poweroff>destroy</on_poweroff>
  <on_reboot>restart</on_reboot>
  <on_crash>destroy</on_crash>

##### ACTUAL RESULTS

VM was creasted with:

  <on_poweroff>destroy</on_poweroff>
  <on_reboot>destroy</on_reboot>
  <on_crash>preserve</on_crash>

thinkdoggie:
##### SUMMARY

Add an optional wait_timeout parameter to community.libvirt.virt_install so the module can wait for multi-phase installs that reboot and continue into a final post-install phase.

The value is specified in seconds, must be a positive integer, and is translated to virt-install --wait rounded up to whole minutes. When omitted, the existing non-blocking behavior is preserved.

Fixes #253

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
`virt_install`

##### ADDITIONAL INFORMATION

Example playbook for an ISO-based unattended install that needs `virt-install` to stay alive across the installer reboot and complete the final post-install phase:

```yaml
- name: Create VM from ISO and wait for unattended install to complete
    name: ubuntu-server
    memory: 2048
    vcpus: 2
    wait_timeout: 1800  # seconds; passed to virt-install --wait as 30 minutes
    disks:
      - size: 25
    osinfo:
      name: ubuntu20.04
    cdrom: /var/lib/libvirt/images/ubuntu-20.04-live-server-amd64.iso
    unattended:
      profile: jeos
      admin_password_file: /tmp/root_password
      user_login: ansible
      user_password_file: /tmp/user_password
    networks:
      - network: default
```

### DIFF
--- a/plugins/modules/virt_install.py
+++ b/plugins/modules/virt_install.py
@@ -34,6 +34,13 @@ options:
     description:
       - Use with present to force the re-creation of an existing VM.
     default: false
+  wait_timeout:
+    type: int
+    version_added: '2.3.0'
+    description:
+      - Maximum time in seconds to wait for the install to complete (rounded up to the nearest whole minute when passed to C(virt-install --wait)).
+      - When unset, the module will not pass C(--wait) to C(virt-install) and will start the install then exit (the module always uses C(--noautoconsole)).
+      - Use this for multi-phase unattended installs that reboot into a final post-install phase.
 extends_documentation_fragment:
     - community.libvirt.virt.options_uri
     - community.libvirt.virt_install.options_memory
@@ -348,6 +355,10 @@ def core(module):
     name = module.params.get('name')
     uri = module.params.get('uri')
     recreate = module.params.get('recreate', False)
+    wait_timeout = module.params.get('wait_timeout')
+    if wait_timeout is not None and wait_timeout <= 0:
+        module.fail_json(
+            msg="wait_timeout must be a positive integer (seconds)")
 
     result = dict(
         changed=False,
@@ -379,7 +390,8 @@ def core(module):
             virtConn.undefine(name)
 
         # run virt-install to create new vm
-        changed, rc, extra_res = virtInstall.execute(dryrun=module.check_mode)
+        changed, rc, extra_res = virtInstall.execute(
+            dryrun=module.check_mode, wait_timeout=wait_timeout)
         result['changed'] = changed
         result['commands'].extend(virtInstall.get_commands())
         result.update(extra_res)
@@ -414,6 +426,7 @@ def main():
             choices=['present', 'absent'],
             default='present'),
         recreate=dict(type='bool', default=False),
+        wait_timeout=dict(type='int'),
     )
 
     # General options
--- a/changelogs/fragments/272-virt-install-wait-timeout.yaml
+++ b/changelogs/fragments/272-virt-install-wait-timeout.yaml
@@ -1,0 +1,2 @@
+minor_changes:
+  - "virt_install - add ``wait_timeout`` parameter to wait for multi-phase unattended installs to complete."
--- a/tests/unit/module_utils/test_virt_install.py
+++ b/tests/unit/module_utils/test_virt_install.py
@@ -2501,6 +2501,43 @@ class TestVirtInstallToolExecute(unittest.TestCase):
         self.assertIn('virt-install', called_args)
         self.assertIn('--noautoconsole', called_args)
 
+    def test_execute_wait_timeout_rounds_up_to_minutes(self):
+        """Test wait_timeout (seconds) is rounded up to whole minutes"""
+        self.mock_module.run_command.return_value = (
+            0, "Domain installation proceeding...", "")
+
+        self.virt_install = VirtInstallTool(self.mock_module)
+        self.virt_install.execute(wait_timeout=90)
+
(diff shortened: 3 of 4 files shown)


thinkdoggie:
@kristvanbesien, @csmart, @Andersson007  Please help review this PR, thanks!

Andersson007 (approved):

plugins/modules/virt_install.py
```diff
@@ -34,6 +34,14 @@
     description:
       - Use with present to force the re-creation of an existing VM.
     default: false
+  wait_timeout:
+    type: int
+    version_added: '2.3.0'
+    description:
+      - Maximum time in seconds to wait for the install to complete.
```
csmart: I think we should combine or re-iterate in this line that the value provided is really rounded up to the nearest whole minute (as the argument that `virt-install` takes), otherwise people sill expect 10 seconds and get 60 and wonder why.

```
      - Maximum time in seconds to wait for the install to complete (rounded up to the nearest whole minute).
```

plugins/modules/virt_install.py
```diff
@@ -34,6 +34,14 @@
...
+      - When unset, the module returns immediately after starting the install.
```
csmart: I could be wrong, but I think when unset this will fallback to the default behaviour, where it waits for the console to close (not necessarily immediate), but it will be immediate if graphics is none (resulting in `noautoconsole`).

I wonder if we should reword that a little bit?

```
      - When unset, the module will not pass --wait argument to virt-install and may return immediately after starting the install.
```

csmart (changes requested):
@thinkdoggie thank you for this! I have no problems with the implementation, it's good - rejecting negative wait values is the right call (even though in `virt-install` it means wait forever, it makes no sense in `ansible` as we need to come back from the task). My only suggestion is to clean up the docs a little bit, because we are taking seconds but converting it to minutes for `virt-install` and I think we just need to make sure that's super clear to the user.

plugins/modules/virt_install.py
thinkdoggie: On the “when unset” wording: in this module we always add `--noautoconsole` to virt-install (see `_build_command()` in plugins/module_utils/virt_install.py). Per the `virt-install(1)` man page, without `--wait` the default is to wait for the console to close, but in the case of `--noautoconsole` it will simply kick off the install and exit. So the “wait for console to close” behavior shouldn’t apply to our invocation.

That said, I agree “returns immediately” is a bit absolute. I’ll reword the docs to something like: when unset, the module does not pass --wait to virt-install and will not wait for the install to complete (it starts the install and exits with --noautoconsole). I’ll also make the seconds→minutes (rounded up) conversion more explicit to avoid confusion.

thinkdoggie:
@csmart，thanks for the careful review! Please review my updates again. Thanks!

Andersson007 (approved):

plugins/modules/virt_install.py
csmart: OK sounds good, thanks

PR #273, opened 2026-06-22
feat(virt_install): add wwn disk parameter

### LINKED ISSUE #271: virt_install: disk `wwn` attribute not supported despite virt-install accepting it
### Summary

The community.libvirt.virt_install module does not expose a wwn key for disk devices, even though the underlying virt-install CLI accepts wwn as a flat --disk suffix. As a result there is no way to set a disk's World Wide Name through the module.

Passing wwn at the disk level → rejected by argument spec validation (Unsupported parameters for (community.libvirt.virt_install) module: disks.wwn).
Passing it as target: { wwn: ... } → forwarded literally as target.wwn, which virt-install does not understand (Unknown --disk parameter 'target.wwn').

serial is supported and works, but wwn is missing from the same disk argument spec.

### Issue Type

Bug Report

### Component Name

virt_install

### Ansible Version

ansible [core 2.21.0]
python version = 3.13.5 (main, May  5 2026, 21:05:52) [GCC 14.2.0]
jinja version = 3.1.6
pyyaml version = 6.0.3 (with libyaml v0.2.5)

### Community.libvirt Version

community.libvirt 2.2.0

### Configuration

_No response_

### OS / Environment

Debian 13

### Steps to Reproduce

- name: Create VM with disk WWN
    name: test-vm
    memory: 2048
    import: true
    osinfo:
      name: debian12
    disks:
      - pool: default
        format: qcow2
        size: 10
        bus: virtio
        wwn: "0x5000c50015ea71ad"

### Expected Results

The module sets the disk's WWN, producing --disk ...,wwn=0x5000c50015ea71ad (the same flat syntax virt-install --disk help advertises), and libvirt renders <wwn> inside the <disk> element.

### Actual Results

fatal: [localhost]: FAILED! => {"changed": false, "msg": "Unsupported parameters for (community.libvirt.virt_install) module: disks.wwn. Supported parameters include: address, backing_format, backing_store, blockio, boot, bus, cache, driver, format, geometry, iotune, path, pool, rawio, readonly, serial, sgio, shareable, size, snapshot, source, sparse, target, transient, transient_opts, vol."}

### Code of Conduct

- [x] I agree to follow the Ansible Code of Conduct

thinkdoggie:
##### SUMMARY
Add `wwn` (World Wide Name) as a top-level parameter in virt_install disk specifications, allowing users to set a disk device WWN via `virt-install --disk wwn=...` (mapping to the wwn element in libvirt domain XML). Documentation is updated and a unit test ensures the parameter is passed through correctly.

Fixes #271

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
`virt_install`

##### ADDITIONAL INFORMATION

Example usage:
```yaml
- name: Install VM with disk WWN
    name: test-vm
    memory: 2048
    disks:
      - size: 20
        bus: scsi
        wwn: "0x5000c50015ea71ad"
```

Expected virt-install argument contains:
`--disk size=20,bus=scsi,wwn=0x5000c50015ea71ad`

### DIFF
--- a/changelogs/fragments/273-wwn-disk-param.yml
+++ b/changelogs/fragments/273-wwn-disk-param.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - virt_install - add ``wwn`` parameter to disk specifications for setting the World Wide Name of a disk device (https://github.com/ansible-collections/community.libvirt/issues/271).
--- a/plugins/doc_fragments/virt_install.py
+++ b/plugins/doc_fragments/virt_install.py
@@ -1049,6 +1049,12 @@ options:
         type: str
         description:
           - An existing libvirt storage volume to use.
+      wwn:
+        type: str
+        description:
+          - World Wide Name for the disk device.
+          - Maps to the V(wwn) element in libvirt domain XML.
+        version_added: 2.3.0
       size:
         type: int
         description:
--- a/plugins/module_utils/virt_install.py
+++ b/plugins/module_utils/virt_install.py
@@ -1509,6 +1509,7 @@ def get_disks_args():
                 path=dict(type='str'),
                 pool=dict(type='str'),
                 vol=dict(type='str'),
+                wwn=dict(type='str'),
                 size=dict(type='int'),
                 sparse=dict(type='bool'),
                 format=dict(type='str'),
--- a/tests/unit/module_utils/test_virt_install.py
+++ b/tests/unit/module_utils/test_virt_install.py
@@ -1087,6 +1087,30 @@ class TestBuildCommand(unittest.TestCase):
         self.assertIn('readonly=no', disk_args[1])
         self.assertIn('shareable=yes', disk_args[1])
 
+    def test_disk_wwn_parameter(self):
+        """Test that wwn parameter is passed through to --disk argument"""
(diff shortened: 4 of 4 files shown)


csmart (commented):
@thinkdoggie this looks pretty good to me, thanks for the contribution. My only thought really is whether we should do any validation on the WWN or just take any string and let libvirt deal with it.

thinkdoggie:
> retty good to me, thanks for the contribution. My only thought really is whether we should do any validation on the WWN or just t

I am inclined to validate the WWW string value by `virt-install`. It's safe to pass-through a string value of disk WWN to the backend `virt-install` tool.

csmart:
Sounds good, do we capture that properly and pass it back to the user?

thinkdoggie:
> Sounds good, do we capture that properly and pass it back to the user?

Yes, that is my understanding as well. For wwn, this follows the module's existing pass-through behavior for virt-install options: we let the backend validate the value, and any resulting error is surfaced back to the user.

I agree the error reporting shape could likely be improved overall, but I think that would be better handled as a separate refactor for the module's command execution error handling rather than something specific to wwn.

